PR #1, opened 2017-03-29
Add typing to requirements

mammuth:
Didn't look deeply in it, but when I wanted to check it out, I noticed that typing was used but isn't listed in your requirements file. 

This PR adds typing (also without version pinning).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ jinja2
 pytest
 requests
 werkzeug
+typing


tomchristie:
The package will by python 3.5+ only, so we should be able to assume that the typing module is built-in, right?

image

tomchristie:
We could consider also targeting Python 3.4 if we can do so without too much effort.

tomchristie:
Also, thanks for the pointer here! Project is still very much in progress, but have things fairly well mapped out, so hoping to progress fairly quickly.

mammuth:
True!

PR #20, opened 2017-04-01
setup test coverage

jairhenrique:


.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+charset = utf-8
+
+[Makefile]
+indent_style = tab
```
tomchristie: Let's drop this. Probably would rather avoid it and it's a seperate concern to the issue at hand in any case.

Makefile
```diff
@@ -0,0 +1,5 @@
+test:
+	PYTHONPATH=. pytest tests
+
+test-coverage:
+	PYTHONPATH=. pytest --cov=apistar --cov-report xml --cov-report term tests
```
tomchristie: Let's use GitHub scripts style instead, adding `/scripts/test`, and `/scripts/test --cov`.

jairhenrique: @tomchristie can I use click?

.editorconfig
jairhenrique: @tomchristie drop all .editorconfig file?

Makefile
tomchristie: Yup!

jairhenrique:
@tomchristie Could you review the new test script?
I also added a configuration file for [codecov](https://docs.codecov.io/docs/codecov-yaml) (https://github.com/tomchristie/apistar/pull/20/commits/1fa3eedc82928e1667913726704a57dd738c90a9)

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+charset = utf-8
```
tomchristie: Let's remove the `.editorconfig` - it's not related to the issue at hand, and I doubt I want to see it included in any case.

scripts/test
```diff
@@ -0,0 +1,21 @@
...
+    if cov:
+        command += ['--cov=apistar', '--cov-report=xml', '--cov-report=term']
+
+    command += ['tests']
+
+    run(command)
+
+
+if __name__ == '__main__':
+    test()
```
tomchristie: Great! 👍 😎 

At some point we can think about improvements here (eg, can we accept arbitrary args on the rest of the command line, and just pass through to py.test as-is?) but those are incremental improvements, and shouldn't block this ticket.

codecov.yml
```diff
@@ -0,0 +1,9 @@
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
+
+  status:
+    project: no
+    patch: yes
+    changes: no
```
tomchristie: I'd love it if there was any way to avoid adding this config file (I wonder if we could write it only during the travis script itself????) or else setup the config in the "coverage" service directly)  But if there's no way around then I'm okay with us including it.

tomchristie: Let's use `.codecov.yml` as the name, instead.
That's also valid, see https://docs.codecov.io/docs/codecov-yaml#section-repository-yaml
That'd be more consistent with our `.gitignore` and `.travis.yml`, and would keep me happy! :smile:. I probably don't mind about the config file existing at that point.

PR #21, opened 2017-04-01
[proposal] code lint

jairhenrique:


Makefile
```diff
@@ -0,0 +1,3 @@
+lint:
+	isort --check-only
+	flake8 .
```
tomchristie: Let's go with GitHub scripts style instead of this `Makefile`. (See https://github.com/tomchristie/apistar/issues/13)

So:

Remove the Makefile, and and a `/scripts/lint` executable command. (Could be either shell script or python, but don't include an extension on the script, and ensure that it's executable)

tomchristie: We'll also want to add `/scripts/test`.

.travis.yml
```diff
@@ -10,4 +10,6 @@ install:
     - pip install -r requirements.txt
     - export PYTHONPATH=.
 
-script: py.test tests
+script:
+  - make lint
+  - py.test tests
```
tomchristie: This'll be:

    /scripts/lint
    /scripts/test

setup.cfg
```diff
@@ -0,0 +1,12 @@
...
+
+[isort]
+atomic=true
+multi_line_output=5
+known_standard_library=types
+known_third_party=pytest
+known_first_party=apistar
+
+[flake8]
+ignore=E501
```
tomchristie: Which commands pick up this config? Is it explicit or implicit?

setup.cfg
```diff
@@ -0,0 +1,12 @@
+[wheel]
+universal = 1
```
tomchristie: Is this needed? Do we already have a universal wheel as part of the package?

setup.cfg
```diff
@@ -0,0 +1,12 @@
...
+
+[isort]
+atomic=true
+multi_line_output=5
+known_standard_library=types
+known_third_party=pytest
+known_first_party=apistar
```
tomchristie: Do we need this `isort` config. How do things differ if we dropped it, and just used the standard? Is it possible to enforce a "everything in one section, alphabetically" policy, rather than the standard "builtin, third-party, first party" sections?

tomchristie:
Planning to pick this up again in a fresh PR, or are you done with this? Let me know either way around, as it's def still worth addressing.

jairhenrique:
@tomchristie I will open new PRs.

PR #24, opened 2017-04-03
readme: syntax-highlighting for python codeblocks

hjwp:
just knocked this together v quickly.  makes gh homepage more colourful.


tomchristie:
Fab. Yes, let's do this!
Thanks!

PR #28, opened 2017-04-06
Allow to specify host and port in run command

### LINKED ISSUE #27: Allow to specify host and port in run command
It should be possible to specify the host and port command for the run command. I would suggest `--host` and `-h` option for host and `--port` and `-p` option for port.

iljabauer:
Resolves #27

### DIFF
--- a/apistar/commands/base.py
+++ b/apistar/commands/base.py
@@ -40,13 +40,15 @@ def new(target_dir, template, force):
 
 
 @click.command(help='Run the current app.')
-def run():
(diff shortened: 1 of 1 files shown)


tomchristie:
Great yup! I'm not going to hold up on the test coverage (looks like it's not configured quite as I want yet. Should be whole repo % only, not patch %) so this is good as-is.

tomchristie:
Great stuff, thanks!

PR #30, opened 2017-04-08
setup flake8 lint script

jairhenrique:
@tomchristie 
What kind of check do you want in flake8? 
Do you prefer a file with the validations or parameters in the command line?

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+ignore=E501
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,6 @@
+[settings]
+atomic=true
+multi_line_output=5
+known_standard_library=types
+known_third_party=pytest
+known_first_party=apistar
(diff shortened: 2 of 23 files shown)


.travis.yml
```diff
@@ -11,7 +11,9 @@ install:
     - pip install codecov
     - export PYTHONPATH=.
 
-script: ./scripts/test --cov
+script:
+  - ./scripts/lint --flake8
```
tomchristie: Ideally Travis would run the Linting as a seperate test run. (Eg see how the Travis builds against django rest framework look)

.isort.cfg
```diff
@@ -0,0 +1,6 @@
+[settings]
+atomic=true
+multi_line_output=5
+known_standard_library=types
+known_third_party=pytest
+known_first_party=apistar
```
jairhenrique: based on `django-rest-framework`.

.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
+ignore=E501
```
jairhenrique: based on `django-rest-framework`

tomchristie:
Ace, thank you!

PR #37, opened 2017-04-10
Use max-line-length = 120

cclauss:
Instead of ignoring all line-too-long errors, set max-line-length to something more reasonable?  127 characters is the width of the GitHub code viewer.


tomchristie:
Have to made a call on this yet, but if we *do* go for it then 120 would be a nice round number.

Issue #2, opened 2017-03-30
Tests on Travis CI


tomchristie:
- [x] Tests running on Travis CI
- [x] Badge

PR #48, opened 2017-04-13
SQL Alchemy Backend

maximilianhurl:
Example usage: https://gist.github.com/maximilianhurl/cb5507354fc8be812c1e3a578627001f

- [x] Move SQL Alchemy initialisation into component
- [x] Add command to create tables
- Add db tests;
    - [x] SQLite
    - [x] PostgreSQL
- [x] Pull config from settings
- [x] Add documentation to readme

Future:

- [ ] Add support or documentation for migrations
- [ ] Use environment settings for db config
- [ ] Add preloaded component (once available)


tomchristie:
👍  👍  👍

maximilianhurl:
Just want to get some feedback on how to implement the initial preloaded component.

1) https://github.com/maximilianhurl/apistar/pull/1 using a naive implementation of preloaded components
2) https://github.com/maximilianhurl/apistar/pull/2 just updating the db component to have a build.

Which would you prefer as the best stop gap? @tomchristie

tomchristie:
Go with `#2` for now.

tomchristie:
You'll probably want to save the `db_engine_config` on the App instance - that way the `create_tables` command will be able to get hold of it, and build the DBBackend.

We'll probably later move that all into a general "settings" state that gets saved on the app instance, but I'd ignore that for now.

tomchristie:
This is looking **✨A⭐️W🌟E⭐️S🌟O⭐️M🌟E✨**.

.gitignore
```diff
@@ -7,4 +7,5 @@ __pycache__
 /htmlcov
 /venv
 /build
+/env
```
tomchristie: Let's prefer the existing `venv`

apistar/routing.py
```diff
@@ -56,8 +56,12 @@ class Router(object):
     }
 
     def __init__(self, routes: List[Route]) -> None:
+        from apistar.app import DBBackend
+        self.not_found = None  # type: RouterLookup
+        self.method_not_allowed = None  # type: RouterLookup
```
tomchristie: These two have been dropped now.

requirements.txt
```diff
@@ -8,5 +8,6 @@ mypy
 pytest
 pytest-cov
 requests
+SQLAlchemy==1.1.9
```
tomchristie: Let's drop the version here for now. If we start pinning requirements then we should do that throughout, rather than just for this one.

At some point we'll probably also need to start breaking this down into eg `# core requirements`, `# optional packages`, `# test requirements`, with SQLAlchemy being an optional. But let's not think about that for now.

tomchristie:
Brill! Want to include a brief example in the ticket?

tomchristie:
Okay, so I've merged in the templates work now so you'll be able to see how we handle settings.

    https://github.com/tomchristie/apistar#templates

You'll want to use something like:

    {
        'DATABASE': {
            'URL': ...,
            'POOL_SIZE': ...,
            'METADATA': ...
        }
    }

tomchristie:
Opened a discussion thread on migrations... http://discuss.apistar.org/t/migrations-with-sql-alchemy/18

We'll tackle that as a separate issue when we get to it, but worth highlighting here.

.travis.yml
```diff
@@ -6,9 +6,19 @@ python:
     - "3.5"
     - "3.6"
 
+env:
+    - DB_URL=sqlite:///test.db
```
maximilianhurl: @tomchristie I've left this is on so it achieves 100% test coverage as there is an exception to allow creation of sqlite engines: https://github.com/tomchristie/apistar/pull/48/files#diff-2c8f3f55d75d3cfd5148140210740646R25

tomchristie:
Brill! I'm hoping to take a look at finishing last few bits of this off, this evening.

Next Q's will be: 1. How do the missing TechEmpower benchmark test types look like with this? 2. Starting thinking about stuff like: there an async option w/ SQLAlchemy? What does sanic recommend for async DB adapters etc.

maximilianhurl:
Feel free to add any comments on things to fix. Should have a bit of time later to look at anything and also make a start on the tech empower benchmarks. Will send over whatever I get done.

2) the sanic examples in the repo all use libs from   https://github.com/aio-libs Which also include support for SQLAlchemy.

tomchristie:
Brill - taking on a last few bits of tweaking now, but about ready to merge.

tomchristie:
---> #104

tomchristie:
Some thoughts in #104 on (1) tweaking project layout slightly, in anticipation of adding other backends in the future, and (2) using 'settings' as the build argument.

tomchristie:
:boom:

PR #52, opened 2017-04-17
add a check to ensure QUERY_STRING actually exists in dictionary

kinabalu:
A possible fix for a missing QUERY_STRING when using gunicorn

### DIFF
--- a/apistar/http.py
+++ b/apistar/http.py
@@ -113,6 +113,8 @@ class Header(str):
 class QueryParams(ImmutableMultiDict):
     @classmethod
     def build(cls, environ: WSGIEnviron):
+	if 'QUERY_STRING' not in environ:
+	    return cls(None)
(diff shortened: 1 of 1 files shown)


tomchristie:
I'm going to merge this in as-is.
A test case to ensure that we don't regress this might not be a bad idea too, but I don't see any reason we shouldn't start off by merging this.

Thank you!

kinabalu:
thought i'd get the change in there first, looking at adding a test case, just wasn't sure where to go about it yet.

tomchristie:
Probably in `test_http`, at the bottom. Doesn't really fit in with anything else.

tomchristie:
On second thoughts a better way to handle this would be to have our test client omit `QUERY_STRING` from the WSGI environ when no query string is present.

tomchristie:
Followed up in #56 - thanks for highlighting this!

PR #91, opened 2017-04-21
Tweak minimal layout

tomchristie:
Considering tweaking the minimal layout slightly, so that the `routes` are created outside of the app itself. Nice thing about that is that it fits in more nicely with when you later also want to add in settings. Eg, this reads nicely...

    routes = [
       ...
    ]
    settings = {
        ...
    }
    app = App(routes=routes, settings=settings)

But this doesn't:

    app = App(routes=[
            ... # Some big routing table
        ], settings={
            ... # Lots of app settings.
        }
    )

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ Create a new project:
     def welcome():
         return {'message': 'Welcome to API Star!'}
 
+
     routes = [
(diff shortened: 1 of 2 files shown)


tomchristie:
Let's 'av it.

Issue #87, opened 2017-04-20
Add missing type annotations


tomchristie:
This would be a great area for contributions. Ie. if you're looking to just start getting involved in the project, then...

* Run `venv/bin/mypy apistar tests --ignore-missing-imports --disallow-untyped-defs`
* Pick one of the modules (in `apistar`, rather than in the `tests` directory, at least for now)
* Go through and see if you can add all the missing type annotations for that module.
* Make a pull request! 🚀

If you do want to take this on, then only add type annotations for a single module in any one pull request. (Rather than making a big pull request which affects multiple modules)

<p align="center">&mdash; ⭐️ &mdash;</p>

In due course we *could* consider adding `--disallow-untyped-defs` to our mypy checking of the project.

This'd give us: `venv/bin/mypy apistar tests --ignore-missing-imports --disallow-untyped-defs`

Doing so would force us to add type annotations *everywhere* within `apistar` itself.
There are some test cases where we deliberately *don't* want to use this, as we're eg. checking functionality that occurs when eg. an untyped view is used, so we'd want to turn it off in those cases. (Or alternatively that might not be something we want to support anyways, and we *enforce* that handled views in API Star are *always* typed?)

So, question:

* What's the comment directive for turning that functionality off on a line-by-line basis? Is there one?

morrme:
@tomchristie does [this](http://mypy.readthedocs.io/en/latest/common_issues.html#spurious-errors-and-locally-silencing-the-checker) help?

weargoggles:
@tomchristie [mypy.ini lets you specify mypy directives on a per-module basis using pattern matches](http://mypy.readthedocs.io/en/latest/config_file.html#example) which seems like it would allow setting `disallow-untyped-defs` for a specific test module.

tomchristie:
@weargoggles That's neat. We might end up wanting to eg. have strict type defs throughout the codebase, but not in the tests.

For now we've still got plenty of outstanding annotations to add in the codebase, before we think about tightening up the policy.

mammuth:
I think most annotations that are easy to catch / don't require deeper code changes are done now. 

Most of the remaining complaints by mypy are in tests or eg. the missing return type of the `build()` methods (#116)

rcshubhadeep:
Taking on the schema module. Will try to add as much type annotations as possible. And maybe will take on some of the the TODOs after the PR related to type annotations are done.

What do you think  @tomchristie ?

tomchristie:
[...]

PR #92, opened 2017-04-21
Add type annotations to main.py

ciplazar:
#87

### DIFF
--- a/apistar/main.py
+++ b/apistar/main.py
@@ -7,16 +7,17 @@ import sys
 
 import click
 
+from apistar.app import App
 from apistar.exceptions import ConfigurationError
 
 sys.dont_write_bytecode = True
 
 
-def get_app_path():
+def get_app_path() -> str:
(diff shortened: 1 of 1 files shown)


apistar/main.py
```diff
@@ -27,18 +28,16 @@ def get_current_app():
     if not hasattr(module, 'app'):
         raise ConfigurationError("The app.py module did not contain an 'app' variable.")
 
-    app = module.app
+    app = module.app  # type: ignore
```
ciplazar: Not sure how else to do this one since mypy complains about `module` not having an `app` attribute otherwise.

tomchristie: How about we use getattr(module, 'app', None) and do 'if app is None' instead of our 'hasattr' check above?

tomchristie:
Great stuff thanks! Have made one suggestion that might be worth trying.

ciplazar:
Nice one!

tomchristie:
Ace, thanks!

PR #97, opened 2017-04-22
Add current directory to `apistar new --layout minimal`

RadoRado:
Otherwise getting:

```
$ apistar new --layout minimal
Usage: apistar new [OPTIONS] TARGET_DIR

Error: Missing argument "target_dir".
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Install API Star:
 
 Create a new project:
 
-    $ apistar new --layout minimal .
+    $ apistar new . --layout minimal
     app.py
     tests.py
     $ cat app.py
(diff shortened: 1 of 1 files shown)


tomchristie:
Yup. Tho let's put the '.' before the '--layout minimal' so it's harder to miss in a cut and paste.

We did used to default to the CWD. We might want to add that behaviour back in?

RadoRado:
@tomchristie You mean something like:

```
$ apistar new $(pwd) --layout minimal
```

tomchristie:
Yes, but just '.' rather than '$(pwd)'

tomchristie:
Having the dot at the very end of the command line means some users might not see it and manage to miss it from a copy n paste (that's my experience from the Django REST framework docs anyhows)

RadoRado:
@tomchristie :+1:

PR #98, opened 2017-04-22
adding missing type annotations for the commands modules #87

kinabalu:
these 3 functions don't return anything, so this annotation should be sufficient.

### DIFF
--- a/apistar/commands/new.py
+++ b/apistar/commands/new.py
@@ -27,7 +27,7 @@ class Force(schema.Boolean):
     default = False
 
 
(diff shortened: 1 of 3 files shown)


tomchristie:
For cases like WSGIResponse.build, the class hasn't finished being declared by the time the annotation is read.

For those cases use a string literal of the class name. (I'd suggest single quotes by convention)

tomchristie:
👍

tomchristie:
Thanks!

PR #100, opened 2017-04-22
adding WSGIResponse with a string literal rather than the class #87

kinabalu:
is this what you meant by using string literal instead?

### DIFF
--- a/apistar/wsgi.py
+++ b/apistar/wsgi.py
@@ -25,7 +25,7 @@ class WSGIResponse(object):
         self.iterator = iterator
 
     @classmethod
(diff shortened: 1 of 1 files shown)


tomchristie:
Ah.

Yes that was exactly what I meant. That's valid python and will make mypy happy, However the failing test is because that doesn't yet work with API Star's type annotation framework.

For now we'll need to not fill in the 'build' return annotations. We'll come back to supporting the literal syntax in due course.

tomchristie:
Closing this for now as it's not something we can handle just yet.

Issue #66, opened 2017-04-20
CORS Support


kinabalu:
Using any API from the browser obviously requires support for CORS.  Django and Flask have several ways to integrate this, and I'd love to see that in this project.  Not sure what the most effective way to enable this is.

In a small API I built just recently, I've got it hardcoded, but it would be great to identify the headers, and turn it on.

tomchristie:
Yes, absolutely!
What headers did you end up setting for your particular case?
What configuration would we want of this?

tomchristie:
See also: CSP

Are there any other security/access response header types that we'll want to consider, too?

kinabalu:
Here are the settings I use on the server-side:

```
Access-Control-Allow-Origin
Access-Control-Allow-Headers
Access-Control-Allow-Methods
Access-Control-Allow-Credentials
Access-Control-Expose-Headers
```

The only other one I know of is `Access-Control-Max-Age` which states how long the browser can cache the preflight.

On the client end the preflight will possibly send `Access-Control-Request-Method` and `Access-Control-Request-Headers`

As far as configuration goes, I'm a fan of a decorator on methods, or something more global that you can add to the routes (since we already know what's there)

For Content-Security-Policy headers, it doesn't seem to be pertinent for API pages? If apistar is only for API calls, I don't know that CSP applies in this case.

tomchristie:
Okay, I'm gonna restrict this to "CORS Support" for now.

However API Star really will be a general purpose framework, just with a primary focus on APIs. (Example, we'll be serving up API docs, which requires us to have both HTML templating and static files support) Yeah we could *potentially* think about alternative names, but it's tricky! 🤔 💭

audiolion:
how about `Tom` ^^

I was a bit confused when you started adding templates, ORM, etc. because `APIStar` made me think it was a faster `rest_framework` that was more loosely coupled and included mypy typings

kinabalu:
Yeah I get that, we're actually using it to serve images for another API we're building, had to drop into WSGI support to achieve that.

How would you go about adding CORS support in an "apistar"-kinda way?

tomchristie:
At the moment, returning 'Response' (You shouldn't need to drop to WSGI now, right?)

One obvious option for now might be introducing an ['HTTP']['DEFAULT_HEADERS'] settings. How does that sound as a starting point?

kinabalu:
Sorry, WSGI was for the images (not CORS), because it seems like apistar wants to utf8 encode the Response.

[...]

PR #101, opened 2017-04-22
showing the problem with importing settings and typing in App #66

kinabalu:
as you can see if I remove the typing in the App.init the tests run.  But that means we are left with no typing for the routes List


tomchristie:
Okay yup, that's fiddly.

PR #102, opened 2017-04-22
attempt adding the annotation types to exceptions class #87

kinabalu:
first crack at this, in order to make it work because of circular dependencies we defer imports in the `schema.py`


tomchristie:
Let's not add any annotations if it means rejigging bits of the codebase.
If there's any simple ones left then let's go for those. We can re-assess any of the fiddly stuff at a later date. 👍

Issue #17, opened 2017-03-30
Redirect handling


tomchristie:
Werkzeug can raise a `RequestRedirect` if eg. view is routed to `/foo/`, and a GET request is made to `/foo` (without the trailing slash).

We ought to be dealing with those in our [exception handler function](https://github.com/tomchristie/apistar/blob/master/apistar/routing.py#L139), and returning an appropriate response.

tomchristie:
Similar to #66. Might be a good thing for someone to tackle first one and then take on the other.

PR #103, opened 2017-04-22
added support for redirect handling in #17 

kinabalu:
handling this by adding the omitted slash in the exception handler. figured this might be better than throwing a 404

### DIFF
--- a/apistar/routing.py
+++ b/apistar/routing.py
@@ -132,6 +132,8 @@ class Router(object):
             raise exceptions.NotFound()
         except werkzeug.exceptions.MethodNotAllowed:
             raise exceptions.MethodNotAllowed()
+        except werkzeug.routing.RequestRedirect:
+            (name, kwargs) = self.adapter.match("{}/".format(path), method)
         (view, pipeline) = self.views[name]
         return (view, pipeline, kwargs)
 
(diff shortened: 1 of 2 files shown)


tomchristie:
Nice idea! However we do actually need to return an HTTP redirect response rather than just routing to the view. We'll want to raise a new kind of exception which also has the URL as a parameter, catch it, and return a "302 Found", with the `Location` header set appropriately.

apistar/routing.py
```diff
@@ -132,6 +132,8 @@ def lookup(self, path, method) -> RouterLookup:
             raise exceptions.NotFound()
         except werkzeug.exceptions.MethodNotAllowed:
             raise exceptions.MethodNotAllowed()
+        except werkzeug.routing.RequestRedirect:
+            (name, kwargs) = self.adapter.match("{}/".format(path), method)
```
tomchristie: Will want to eg. `raise exceptions.Found(url)`

tests/test_routing.py
```diff
@@ -89,6 +89,14 @@ def test_405():
     }
 
 
+def test_found_no_slash():
+    response = client.get('/found')
+    assert response.status_code == 200
```
tomchristie: We'd like a 302, with the `Location` header set.

PR #105, opened 2017-04-24
SQLAlchemy tweaks

tomchristie:


apistar/backends/sqlalchemy.py
```diff
@@ -16,7 +16,7 @@ def build(cls, settings: Settings):
         metadata = config.get('METADATA')
 
         kwargs = {}
-        if url.startswith('postgresql:'):  # pragma: no cover
+        if url.startswith('postgresql'):  # pragma: no cover
```
maximilianhurl: This probably wants to be `if not url.startswith('sqlite'):` as mysql may also want to set the pool size

audiolion: what if someone is using `mongodb`? then that would cause a bug, maybe explicitly listing if statements is better?

tomchristie: Idealt we have a set of all the values that we *do* want to include pool size on.

PR #110, opened 2017-04-25
Type-annotate apistar.exceptions

weargoggles:


tomchristie:
👍  Ta!

tomchristie:
Okay I'm clear not paying enough attention, actually the messages can be dicts too. Not obv. why mypy hasn't picked up on that.

tomchristie:
Changing the name to `detail` now, as I think that's marginally less deceptive (ie. exception details can also be data structures)

weargoggles:
Needs to be run with --check-untyped-defs to check calls to already typed functions inside those not yet typed.

tomchristie:
Good call! Looking into the output of that now!

tomchristie:
Ran into some roadblocks trying to switch that flag on, so I've punted on it for now.

PR #113, opened 2017-04-25
Redirect handling.

### LINKED ISSUE #17: Redirect handling

kinabalu:
Return HTTP 302 responses when we get a redirect from Werkzeug.

Closes #17

### DIFF
--- a/apistar/routing.py
+++ b/apistar/routing.py
@@ -20,7 +20,8 @@ primitive_types = (
 )
 
 schema_types = (
-    schema.String, schema.Integer, schema.Number, schema.Boolean
+    schema.String, schema.Integer, schema.Number, schema.Boolean,
+    schema.Enum, schema.Object
 )
 
 Route = namedtuple('Route', ['path', 'method', 'view'])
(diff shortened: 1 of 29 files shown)


apistar/routing.py
```diff
@@ -138,11 +138,17 @@ def lookup(self, path: str, method: str) -> RouterLookup:
             raise exceptions.NotFound()
         except werkzeug.exceptions.MethodNotAllowed:
             raise exceptions.MethodNotAllowed()
+        except werkzeug.routing.RequestRedirect:
+            raise exceptions.Found(path + '/')
```
tomchristie: Does `RequestRedirect` include something like a url property already?
Can we do, something like this maybe?...

    except werkzeug.routing.RequestRedirect as exc:
        raise exceptions.Found(exc.url)

tests/test_routing.py
```diff
@@ -89,6 +89,14 @@ def test_405():
     }
 
 
+def test_found_no_slash():
+    response = client.get('/found')
+    assert response.status_code == 200
```
tomchristie: It'd probably make more sense for us to check the result of `client.get('/found', follow_redirects=False)`, and ensure that a 302 is returned, with the correct headers.

tomchristie:
(updated title)

Also, apologies looks like I resolved a merge conflict before you were done working on the PR. GitHub just made it too nice and easy! 😄

kinabalu:
heh, easy enough to fix the conflict, just like we likes it

tomchristie:
Ace, thanks!

PR #114, opened 2017-04-26
Type-annotate apistar.wsgi

mammuth:
When we want to annotate the return type of the `build` classmethods [PEP-484](https://www.python.org/dev/peps/pep-0484/#forward-references) suggest a string literal annotation like 
_... -> 'WSGIResponse'_

The tests showed, that this breaks `pipelines.get_class_id` so I also added a small change to fix this.

Is this the proper way to annotate the build classmethods and does it make sense?

### DIFF
--- a/apistar/pipelines.py
+++ b/apistar/pipelines.py
@@ -41,7 +41,7 @@ def parameterize_by_argument_name(cls: type) -> bool:
 def get_class_id(cls: type,
                  arg_name: str=None) -> str:
     # http://stackoverflow.com/questions/1175208/elegant-python-function-to-convert-camelcase-to-snake-case
-    name = cls.__name__
+    name = cls if type(cls) is str else cls.__name__
     s1 = FIRST_CAP_RE.sub(r'\1_\2', name)
     s2 = ALL_CAP_RE.sub(r'\1_\2', s1).lower()
     if parameterize_by_argument_name(cls):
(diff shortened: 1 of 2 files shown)


tomchristie:
Great stuff!

> Is this the proper way to annotate the build classmethods

Yes it is, absolutely. We just need to make sure that it works properly with our dependency resolution. Wondering if adding the `build` return annotations to the `http` module might not be a bad idea as part of this PR, to give us confidence that it all works as expected.

apistar/pipelines.py
```diff
@@ -41,7 +41,7 @@ def parameterize_by_argument_name(cls: type) -> bool:
 def get_class_id(cls: type,
```
tomchristie: Worth noting that `cls: type` isn't correct here anymore (since it could also be a string).
Don't know if there's already an annotation that refers to "the type of a type annotation", but otherwise we should probably be using something like `Union[type, str]` here.

Also wonder if that'd allow us to remove the odd `str(cls)` below? We shouldn't need to do that.

mammuth:
> Wondering if adding the build return annotations to the http module might not be a bad idea as part of this PR, to give us confidence that it all works as expected.

Tested adding the string type annotations to the http module. This gives us confidence, that it all breaks :grin: 
Well, not all, but It introduces some key-errors in `app.get_wsgi_server`, as seen in this [test log](https://pastebin.com/BLVdXzed).

tomchristie:
Well, glad we checked! 😄 

Nonetheless it's a start. I think we should probably punt on this for now, and I'll put support for string annotation style onto the backlog as a fully fledged issue.

Thanks for helping progress things!

mammuth:
Great :+1:

PR #127, opened 2017-04-28
Type annotate apistar.exceptions

mammuth:
kwargs could probably be annotated more concrete than `Any`, eg. `Union[str, dict, int]` (since that are the currently only possible kwargs in APIException), however this raises the following errors:

```
apistar/exceptions.py:31: error: Argument 2 to "__init__" of "APIException" has incompatible type **Dict[str, Union[str, Dict[Any, Any], int]]; expected "Union[str, Dict[Any, Any]]"
apistar/exceptions.py:31: error: Argument 2 to "__init__" of "APIException" has incompatible type **Dict[str, Union[str, Dict[Any, Any], int]]; expected "int"
```
This seems related to an issue with mypy: https://github.com/python/mypy/issues/1969

That's why I went with `Any` for the kwargs. I hope this is helpful 😊 

Related issue #87

### DIFF
--- a/apistar/exceptions.py
+++ b/apistar/exceptions.py
@@ -1,4 +1,4 @@
-from typing import Any, Union
+from typing import Union
 
 
 class ConfigurationError(Exception):
@@ -26,9 +26,9 @@ class Found(APIException):
     default_status_code = 302
     default_detail = ''
 
-    def __init__(self, location: str, *args: Any, **kwargs: Any) -> None:
+    def __init__(self, location: str, detail: Union[str, dict]=None, status_code: int=None) -> None:
         self.location = location
-        super().__init__(*args, **kwargs)
+        super().__init__(detail, status_code)
 
 
 class ValidationError(APIException):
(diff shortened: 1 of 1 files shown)


apistar/exceptions.py
```diff
@@ -26,7 +26,7 @@ class Found(APIException):
     default_status_code = 302
     default_detail = ''
 
-    def __init__(self, location, *args, **kwargs):
+    def __init__(self, location: str, *args: Any, **kwargs: Any) -> None:
```
tomchristie: I'd suggest that we get rid of `*args, **kwargs`, and explicitly use `detail` and `status_code` instead.

tomchristie:
Great 👍 .

For annotated functions/methods, I reckon we should also move to this style consistently:

    def foo(self,
                 something: Stuu
                 another: Annotation) -> Blah

But welcome to tackle that separately.

mammuth:
This doesn't look so nice when there's only one argument, in my opinion. What you wanna do in this case?
```
def foo(self,
        something: Stuu) -> Blah:
```

tomchristie:
Slightly undecided. Possibly do it for all annotated signatures always, or possibly don't worry about short ones.

PR #128, opened 2017-04-28
Add line breaks in signature for annotated methods

mammuth:
This PR is a response to the comments at https://github.com/tomchristie/apistar/pull/127#issuecomment-298121809. 
I think I reformatted all type annotated methods in the project in regards to this style guideline 🎉 
I also broke signatures with only one annotated argument, so we could have the guideline be simple like "always linebreak annotated signatures" 

I created multiple commits for the different modules in case we want to revert something. One commit additionally adds a missing type annotation in the settings module (I added it since I already touched that line).

### DIFF
--- a/apistar/exceptions.py
+++ b/apistar/exceptions.py
@@ -6,8 +6,7 @@ class ConfigurationError(Exception):
 
 
 class SchemaError(Exception):
-    def __init__(self,
-                 detail: Union[str, dict]) -> None:
+    def __init__(self, detail: Union[str, dict]) -> None:
         self.detail = detail
         super().__init__(detail)
 
--- a/apistar/exceptions.py
+++ b/apistar/exceptions.py
@@ -6,7 +6,8 @@ class ConfigurationError(Exception):
 
 
 class SchemaError(Exception):
-    def __init__(self, detail: Union[str, dict]) -> None:
+    def __init__(self,
+                 detail: Union[str, dict]) -> None:
         self.detail = detail
         super().__init__(detail)
 
@@ -17,7 +18,9 @@ class APIException(Exception):
     default_status_code = 500
     default_detail = 'Server error'
 
-    def __init__(self, detail: Union[str, dict]=None, status_code: int=None) -> None:
+    def __init__(self,
+                 detail: Union[str, dict]=None,
+                 status_code: int=None) -> None:
         self.detail = self.default_detail if (detail is None) else detail
         self.status_code = self.default_status_code if (status_code is None) else status_code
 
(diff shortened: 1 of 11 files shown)


tomchristie:
Okay, after being able to see the side-by-side diff, I'm less convinced by my previous comment.
Perhaps let's only start with this style for >80 char lines.
Apologies for prompting you into unnecessary work!

apistar/exceptions.py
```diff
@@ -17,7 +18,9 @@ class APIException(Exception):
     default_status_code = 500
     default_detail = 'Server error'
 
-    def __init__(self, detail: Union[str, dict]=None, status_code: int=None) -> None:
+    def __init__(self,
```
tomchristie: This one looks good.

apistar/exceptions.py
```diff
@@ -26,7 +29,10 @@ class Found(APIException):
     default_status_code = 302
     default_detail = ''
 
-    def __init__(self, location: str, detail: Union[str, dict]=None, status_code: int=None) -> None:
+    def __init__(self,
```
tomchristie: Yup.

apistar/http.py
```diff
@@ -196,20 +215,29 @@ def build(cls, data: RequestData, arg_name: ArgName):
...
+              url: URL,
+              headers: Headers):
         return cls(method=method, url=url, headers=headers)
 
 
 class Response(object):
     __slots__ = ('data', 'content', 'status', 'headers')
 
-    def __init__(self, data: Any, status: int=None, headers: HeadersType=None) -> None:
+    def __init__(self,
```
tomchristie: Yup.

apistar/routing.py
```diff
@@ -53,7 +55,9 @@ def build(cls, args: URLPathArgs, arg_name: ArgName):
 
 
 class Router(object):
-    def __init__(self, routes: List[Route], initial_types: List[type]=None) -> None:
+    def __init__(self,
```
tomchristie: Yup.

apistar/routing.py
```diff
@@ -158,7 +164,8 @@ def lookup(self, path: str, method: str) -> RouterLookup:
         return (view, pipeline, kwargs)
 
 
-def exception_handler(environ: wsgi.WSGIEnviron, exc: Exception) -> http.Response:
+def exception_handler(environ: wsgi.WSGIEnviron,
```
tomchristie: Seems reasonable, yes.

apistar/wsgi.py
```diff
@@ -19,13 +19,17 @@
 class WSGIResponse(object):
     __slots__ = ('status', 'headers', 'iterator')
 
-    def __init__(self, status: str, headers: List[Tuple[str, str]], iterator: Iterable[bytes]) -> None:
+    def __init__(self,
```
tomchristie: Yup.

mammuth:
Alright, removed the linebreaks on short signatures. It really looked strange for short signatures 👍

tomchristie:
Ace, thanks @mammuth!

PR #138, opened 2017-05-09
Database Backend

ahopkins:
The attempt here is to *begin* to abstract away the database backends and ORMs. The main addition is a new method in `backends.db_backend_loader` that reads a string in the `settings`. It is set to load the existing `SQLAlchemy` backend by default.

Therefore, if your settings were (as in the example):

    settings = {
        "DATABASE": {
            "URL": "postgresql://:@localhost/apistar",
            "METADATA": Base.metadata
        }
    }

It would be implied as:

    settings = {
        "DATABASE": {
            "URL": "postgresql://:@localhost/apistar",
            "METADATA": Base.metadata,
            "BACKEND": "apistar.backends.sqlalchemy.SQLAlchemy"
        }
    }

The `SQLAlchemy` backend should otherwise function as existing. There are two new class level variables (`preload_key` and `commands`) to accomplish the same functionality as already existing in `App()`. Perhaps there will be a better method of handling down the road.

Lastly, there is the beginning of a `BaseDatabaseBackend` that can be used for further abstraction down the road.

One thing to note ... I **did** have to remove a type annotation in `backends.sqlalchemy.SQLAlchemy`. In the `build` method, I had to remove the annotation for the `settings` parameter because there was an instantiation of `settings` problem where the backend was being called before the `app.App` was created.

### DIFF
--- a/apistar/app.py
+++ b/apistar/app.py
@@ -5,7 +5,8 @@ from typing import Any, Callable, Dict, List
 import click
 
 from apistar import commands as cmd
-from apistar import pipelines, routing, schema
+from apistar import pipelines, routing, schema, backends
+# from apistar.backends.utils import db_backend_loader
 
 DEFAULT_LOOKUP_CACHE_SIZE = 10000
 
@@ -25,7 +26,6 @@ class App(object):
         from apistar.settings import Settings
         from apistar.statics import Statics
         from apistar.templating import Templates
-        from apistar.backends.sqlalchemy import SQLAlchemy
 
         routes = [] if (routes is None) else routes
         commands = [] if (commands is None) else commands
@@ -42,9 +42,10 @@ class App(object):
             initial_types.append(Templates)
             self.preloaded['templates'] = Templates.build(self.settings)
         if 'DATABASE' in self.settings:
-            initial_types.append(SQLAlchemy)
-            self.preloaded['sql_alchemy'] = SQLAlchemy.build(self.settings)
-            self.commands += [cmd.create_tables]
+            Backend = backends.utils.db_backend_loader(self.settings)
+            initial_types.append(Backend)
+            self.preloaded[Backend.preload_key] = Backend.build(self.settings)
+            self.commands += Backend.commands
         if 'STATICS' in self.settings:
             initial_types.append(Statics)
             self.preloaded['statics'] = Statics.build(self.settings)
--- a/apistar/app.py
+++ b/apistar/app.py
@@ -5,7 +5,7 @@ from typing import Any, Callable, Dict, List
 import click
 
 from apistar import commands as cmd
-from apistar import pipelines, routing, schema, backends
+from apistar import backends, pipelines, routing, schema
 # from apistar.backends.utils import db_backend_loader
 
 DEFAULT_LOOKUP_CACHE_SIZE = 10000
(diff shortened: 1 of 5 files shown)


tomchristie:
Great stuff, thanks so much for taking a start at this! 😄 
So, I've got a slightly different approach to this in mind, which I'll talk through here...

Rather than have an explicit set of components that the App handles preloading on App.__init__, we'll instead allow components to specify `preload = True` on the class. Then during app init, we'll walk through all the routes and commands, and build any component that's listed as preloading.

With that in mind I'd consider approaching this the other way around... Build out a PeeWee component, ignoring the preloading aspect for now, and just work on the basis that I'll cover that part of the problem.

At least for the moment we won't *have* a generalized abstraction for database backends, because they all expose different enough APIs that you can't just trivially interchange them. But we *will* have a decentralized approach where the application doesn't need to *know* anything about the particulars of the installed database backend - it just automatically gets picked up and loaded during initialization, because it's routed to somewhere in the application. (*)

There's one other minor part to that puzzle, which is making sure that the right set of commands are available to the application, without the developer having to explicitly include them. For that we'll likely have something similar to Flask's blueprints, so that there's a shortcut for registering a bunch of different commands, routes, template dirs etc... However I think that's something we can gloss over for now.

One other thing worth noting is that at some point we'll want to add support for context-manager components. That'd allow use to use eg `peewee_backend.Transaction` or `sqlalchemy.Session` and have commit/rollback handled automatically. Again, that's something to consider for the future. For now we'd just want to expose the `db` instance, and have the developer explicitly use `with db.transaction() as txn:` inside the view.

I'm *very* keen on helping out wherever possible here, so please do let me know *any* questions you have about any of the above. The final thing worth noting is that I'm super keen that we end up pushing alternative backends out as separate third-party packages, rather than bundling everything into the main repo itself. Totally happy for any work to start off as pull requests directly against this repo tho, as that's a perfectly reasonable point from which to progress.

(*) The place where we *will* end up having generalized abstractions will be cases that can have a common interface that applies to more that one backend. Template rendering (get_template, render), Cache backends (get/set) are examples of cases like those. *Perhaps* there's also some very minimal amount of interface that's common across database ORMs, but if there is I'm happy for us to cross that bridge at a later date.

ahopkins:
Awesome. I was thinking the same that it sort of would be a weird abstraction. I knew this wasn't exactly the right solution, but wanted to hear exactly your feedback like this. 

I'll take this idea and try to start running with it.

ahopkins:
Also, I had been planning on making the peewee component a 3rd party package. Maybe, with the same logic the SQLAlchemy module also at some point should be broken off. 

Another thought I had too though, was as much as possible, it would make sense to try and keep the commands consistent, and, for example, reuse things like `create_table` just execute the same regardless of the ORM.

tomchristie:
> I'll take this idea and try to start running with it.

Brill! 👍  👍  👍 

> as much as possible, it would make sense to try and keep the commands consistent, and, for example, reuse things like create_table just execute the same regardless of the ORM.

Agreed, yup.

> Maybe, with the same logic the SQLAlchemy module also at some point should be broken off.

Possibly so. Most likely I'd take a similar approach as have done with REST framework, and make sure that for every big component type we've got at least one built-in option that we guarantee to support as the core functionality, and a range of third-party options, owned by their own maintainers. However, we can take a view on all that as things progress.

tomchristie:
As I say, feel free to either start of by issuing your work directly as a pull request against this repo, or else just start out with a third party package in the first place and go from there. Whichever you prefer.

tomchristie:
Preloading is getting into place now... https://github.com/tomchristie/apistar/pull/139

Add `preload = True` to the component class, and it'll be initialized once only, during `app.__init__`

PR #144, opened 2017-05-10
Use twine for publishing distributions to PyPI

hackebrot:
Hi 👋 

I noticed that API Star uses ``python setup.py upload`` for publishing distributions to PyPI. This method is not recommended due to its security flaws, please see [PyPA guide](https://packaging.python.org/distributing/#uploading-your-project-to-pypi).

This PR replaces said method with ``twine upload dist/*``.

Please let me know your thoughts! 🙇

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,6 @@ werkzeug
 # Optional
 psycopg2
 sqlalchemy
-twine
 whitenoise
 
 # Testing requirements
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,7 @@ werkzeug
 # Optional
 psycopg2
 sqlalchemy
+twine
 whitenoise
 
 # Testing requirements
--- a/scripts/publish
+++ b/scripts/publish
@@ -9,7 +9,8 @@ fi
 find apistar -type f -name "*.py[co]" -delete
 find apistar -type d -name __pycache__ -delete
 
-${PREFIX}python setup.py sdist upload
+${PREFIX}python setup.py sdist
+${PREFIX}twine upload dist/*
 
 rm -r dist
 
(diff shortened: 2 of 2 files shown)


tomchristie:
Sounds good yup! 😄

requirements.txt
```diff
@@ -9,6 +9,7 @@ werkzeug
 # Optional
 psycopg2
 sqlalchemy
+twine
```
tomchristie: Since we only need this for pushing to PyPI, let's not put this in the `requirements.txt`.

We could instead either:

1. Error in `scripts/publish` if the binary file doesn't exist.
2. `pip install twine` in `scripts/publish` if the binary file doesn't exist.

hackebrot: That's a good point! 👍 I just pushed a command to check for the ``twine`` cmd and error out if it cannot be found.

I wouldn't like scripts to install dependencies automatically. The risk of not having a virtualenv active is maybe a bit too high. 🙊

tomchristie:
Fab! Thank you, great catch. 🌟

hackebrot:
Thanks for merging this 😄 

As a side note: I used ``command -v`` because it will check if ``twine`` is on the path. I personally use virtualenvwrapper, which means my virtualenv is in a different location than the publish script assumes. In other words [``[ ! -f ${PREFIX}twine ]``](https://github.com/tomchristie/apistar/blob/master/scripts/publish#L9) will report an error even though twine is installed in my env.

tomchristie:
Ah good call! I've now tweaked that to something in between both of ours.
I *don't* activate the virtualenv when working with apistar locally, but instead just explicitly make sure the use the venv binaries. I think how it's changed to now should cover both cases:

https://github.com/tomchristie/apistar/blob/master/scripts/publish

hackebrot:
Yay! Thank you 👍

PR #146, opened 2017-05-10
Ignore '__pycache__' directories for new command

### LINKED ISSUE #145: Command 'new' copies *.pyc files
Hi there! 👋 

Thank you for making API Star 😍 🚀 

While playing around with the CLI, I noticed that ``apistar new`` copies ``*.pyc`` files.

```
$ pip install apistar
$ apistar new .
./app.py
./__pycache__/app.cpython-36.pyc
./project/__init__.py
./project/routes.py
./project/views.py
./project/__pycache__/__init__.cpython-36.pyc
./project/__pycache__/routes.cpython-36.pyc
./project/__pycache__/views.cpython-36.pyc
./tests/test_app.py
./tests/__pycache__/test_app.cpython-36.pyc
```

Even though there's only a sdist distribution on PyPI, you will see this issue with newer versions of pip due to it's local caching in which it attempts to build ``bdist_wheels`` locally.

> When no wheels are found for an sdist, pip will attempt to build a wheel automatically and insert it into the wheel cache.

Please see [pip docs](https://pip.pypa.io/en/stable/reference/pip_install/#wheel-cache)

You can replicate this issue with installing from PyPI in a fresh virtualenv or by installing from a local wheel distribution and then running ``$ apistar new`` with any layout option.

```
$ python setup.py bdist_wheel
$ pip install dist/apistar-0.1.17-py3-none-any.whl
```

hackebrot:
Resolve #145 

This is slightly awkward to test since it only happens for ``bdist_wheel`` distributions, I'm afraid 😞 

Please let me know how you would like to cover this problem in the unit tests, if at all, and I'll happily extend this PR. 🙇

### DIFF
--- a/apistar/commands/new.py
+++ b/apistar/commands/new.py
@@ -10,6 +10,7 @@ from apistar import schema
 APISTAR_PACKAGE_DIR = os.path.dirname(apistar.__file__)
 LAYOUTS_DIR = os.path.join(APISTAR_PACKAGE_DIR, 'layouts')
 LAYOUT_CHOICES = os.listdir(LAYOUTS_DIR)
+IGNORED_DIRECTORIES = ['__pycache__']
 
 
 class TargetDir(schema.String):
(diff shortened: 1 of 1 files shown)


tomchristie:
Great thanks! I was omitting this on the basis that the layout directories *shouldn't* ever include `__pycache__` directories in the first place, but obviously there are cases where they do.

hackebrot:
Always, man. Keep up the great work! ⭐️ 🚀 ⭐️

PR #147, opened 2017-05-11
Fixed Product schema in README

brosner:
I came to learn the documentation was incorrect and did not work for me. After digging in further I found out that `schema.Object` properties need to be in the `properties` dict and not defined on the class directly.


tomchristie:
Oh my. Yes, thank you!

PR #153, opened 2017-05-17
Django ORM Support

linuxlewis:
Overview
---

This is a first draft of adding support for Django ORM. I have a working project example in a separate repo [here](https://github.com/linuxlewis/djangostar/tree/master/djangostar/example_project).

The idea for this component is that is returns an "models" object to which one accesses their models (after they've been loaded appropriately by django).

```python
def my_view(models: DjangoBackend):
     Star = models.Star
     s = Star(name="My Star")
     s.save()
     return {'name': star.name, 'id': star.id}
```

To configure the backend it expects 2 existing django settings: [DATABASES](https://docs.djangoproject.com/en/1.11/ref/settings/#databases) and [INSTALLED_APPS](https://docs.djangoproject.com/en/1.11/ref/settings/#installed-apps)
An example configuration can be found [here](https://github.com/linuxlewis/djangostar/blob/master/djangostar/example_project/app.py#L4-L15)

Any feedback or questions are welcomed.

Questions
---
- Can components have an API to register commands?
  - This seems to be the main blocker for this PR. I have a few ideas of how it could be done but I would defer to you if you have a implementation in mind.

### DIFF
--- a/apistar/app.py
+++ b/apistar/app.py
@@ -36,7 +36,7 @@ class App(object):
             'app': self
         }
         preload_state(self.preloaded, self.routes)
-        if 'sql_alchemy' in self.preloaded:
+        if 'django_backend' in self.preloaded:
             self.commands += [cmd.create_tables]
 
         self.router = routing.Router(self.routes, initial_types)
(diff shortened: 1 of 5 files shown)


tomchristie:
This is awesome, thanks so much!

@nirgn975 has picked this up in the PyCon sprints and progressed it a little further, including your initial commits. Thank you!!! 🌟

linuxlewis:
👍

Issue #154, opened 2017-05-17
README is not complete when talking about SQLAlchemy


kinabalu:
There are a few extra steps needed before you can get access to SQLAlchemy and the `create_tables` command due to the preloader.

tomchristie:
Our README instructions now include *explicitly* adding the commands to the app,
so we can close this off.

In the future we may have something like blueprints/installed_apps that handles additional app instantiation stuff based on a set of installed packages.

PR #155, opened 2017-05-17
adds more description and linear thought to the SQLAlchemy integration #154

kinabalu:
this is a proposed change to the readme which would solve #154


nirgn975:
@tomchristie I think we can close this one because I [changed](https://github.com/tomchristie/apistar/pull/168/files#diff-04c6e90faac2675aa89e2176d2eec7d8R610) the README to include the commands explicitly.

tomchristie:
Concur. Including the commands explicitly is a good call for the moment.

PR #162, opened 2017-05-22
Minor type annotation tweaks across these files.

tadhg-ohiggins:
Mostly adding function annotations.

### DIFF
--- a/apistar/commands/create_tables.py
+++ b/apistar/commands/create_tables.py
@@ -1,12 +1,12 @@
 import click
 
 
-def create_tables():
+def create_tables() -> None:
     """
     Create SQLAlchemy tables.
     """
     from apistar.main import get_current_app
-    from apistar.backends import SQLAlchemy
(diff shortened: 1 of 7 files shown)


apistar/commands/create_tables.py
```diff
@@ -1,12 +1,12 @@
...
 
 
-def create_tables():
+def create_tables() -> None:
     """
     Create SQLAlchemy tables.
     """
     from apistar.main import get_current_app
-    from apistar.backends import SQLAlchemy
+    from apistar.backends.sqlalchemy import SQLAlchemy
```
tadhg-ohiggins: Even though all the tests pass with or without the `.sqlalchemy` tail, MyPy complains about it if I run `venv/bin/mypy apistar --ignore-missing-imports --disallow-untyped-defs` locally and `.sqlalchemy` is absent.

tomchristie:
Nice!

Issue #77, opened 2017-04-20
Routing


tomchristie:
Still some bits to do:

* URL Reversing
* Support `include`, and create sections in API schemas/docs.
* Figure out hostname binding (eg. search for example.com in our codebase)
* Path components
* Remove `URLPathArg` guard if `URLPathArgs` is used.

sandys:
DRF has a very public stand against nested routing. I'm not sure what is wrong or right, but for people *migrating* to a new framework (like Apistar), nested routes is a must have. Especially when you have live mobile apps, it is impossible to correct mistakes in early API design. 

Would you support nested routes in apistar?

tomchristie:
The difficulties with nested routing in REST framework are more around how the generically provided bits of behavior (ie. the generic views) would work in that case, and also around how you handle automatic hyperlinking in that case.

We won't be planning on bringing in quite so much automagical generic behavior into API Star, so yes you're absolutely free to structure the API however you wish.

There is one thing related to this where API Star will be able to do a better job than REST framework currently does, which is about how nested URLs get handled in schema generation, schema layout, and API docs. That's difficult in REST framework since we have to automagically infer what schema layout would work well. In API Star that'll be built into the structure of the application itself. (Eg. using "includes" to create new sections within a schema)

sandys:
This is awesome!
Is the nested routes currently available or is it a future plan ? Only reason I ask is because your examples are the simpler, single level routes.

tomchristie:
We don't have 'includes' but you can structure the URL layout however you want, yes.

tomchristie:
We have all the functionality listed here now.

PR #164, opened 2017-05-22
Support for URL Reversing in Router

dorianpula:
Add support for [Flask-like `url_for()` URL building](http://flask.pocoo.org/docs/0.12/api/#flask.url_for) in API Star:

# Details
- Returns a relative URL.
- Part of work for issue #77

### DIFF
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -226,21 +226,21 @@ def test_lookup_cache_expiry():
         assert response.json() == {'path': '/%d/' % index}
 
 
-@pytest.mark.skip('WIP')
 def test_routing_reversal_on_path_without_url_params():
-    pass
+    url = app.router.reverse_url('found')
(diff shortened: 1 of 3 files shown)


tests/test_routing.py
```diff
@@ -224,3 +224,23 @@ def get_path(var: int, path: http.Path):
...
         assert response.json() == {'path': '/%d/' % index}
+
+
+def test_routing_reversal_on_path_without_url_params():
+    url = app.router.reverse_url('found')
+    assert url == 'http://localhost:8080/found/'
+
+
+def test_routing_reversal_on_path_non_existent_path():
+    with pytest.raises(exceptions.NotFound):
```
tomchristie: We'll have something like a `NoReverseMatch` (resulting in a 500 error), rather than `NotFound` (which would imply a 404 response)

PR #165, opened 2017-05-22
Update the README to use the Rating class within Schemas

kristenwidman:
In the Schemas section of the README, a class called Rating is defined, but is not used within the Product class (which instead defines another identical rating with integers). This change is just to use that class rather than redefining it.


tomchristie:
Thanks! 👍

PR #167, opened 2017-05-22
Removes static server name since it's no longer required.

Sonictherocketman:
Redirects are still functional without this hardcoded `example.com`. 

[RFC 7231 &#8594;](https://tools.ietf.org/html/rfc7231#section-7.1.2)
[Werkzueg Router Docs &#8594;](http://werkzeug.pocoo.org/docs/0.12/routing/#werkzeug.routing.Map.bind)


tomchristie:
😎

PR #169, opened 2017-05-22
Add support for converting to coreschema types.

jriggins:
Currently Boolean, Integer, and Number types

### DIFF
--- a/apistar/apischema.py
+++ b/apistar/apischema.py
@@ -52,6 +52,17 @@ def get_schema_content(routes: RoutesConfig) -> Dict[str, Route]:
     return content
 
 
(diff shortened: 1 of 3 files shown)


jriggins:
image

image

image

tomchristie:
💥

adawolfs:
What is the main purpose of this? I'm looking for a way to implement [**this**](https://github.com/tomchristie/apistar/issues/211) and I've found that it's caused because the defined types are replaced by coreapi types

tomchristie:
This ensures that we have rich type descriptions in the schema, and can render appropriate types of controls. Later we can include descriptions too, which we're not yet doing.

PR #170, opened 2017-05-22
Allow forward references (str-type) in annotations

maackle:
Allows strings as type annotations in order to break circular references.

This works fine with mypy static type checking as long as the module is included. For instance:

```py
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    # if imported at runtime, this will cause a circular import,
    # but mypy needs it in this module in order to resolve it
    import apistar.settings

class A:
    @classmethod
    def build(cls, settings: 'apistar.settings.Settings'):
        # ...
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .cache
 .coverage
 .git
+.mypy_cache
 __pycache__
 /*.egg-info
 /htmlcov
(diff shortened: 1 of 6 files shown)


tomchristie:
Radical! I want to mull on this for a bit before pull it in, but seems legit.

tomchristie:
The work in #235 means that we'll end up with a nice separation between interfaces and implementations of those interfaces. There's no circular dependencies left and it's not clear that we'll have any great need to support this now. Might be that it comes up again in the future, but as things stand I'm going to close this off. Thanks so much for stepping in - might be that the work comes in useful later down the line.

PR #173, opened 2017-05-23
Add `@builder` decorator as alternative to `build()` classmethod

maackle:


maackle:
The linter is failing in CI but runs fine locally, not sure why.

apistar/commands/schema.py
```diff
@@ -22,7 +22,7 @@ def schema(format: Format) -> None:  # pragma: nocover
     from apistar.cli import get_current_app
     from apistar.apischema import APISchema
     app = get_current_app()
-    schema = APISchema.build(app)
+    schema = core.get_builder(APISchema)(app)
```
tomchristie: Just use `build_apischema` directly here, for now.

apistar/settings.py
```diff
@@ -21,11 +18,14 @@ def get(self, indexes: Union[str, List[str]], default: Any=None) -> Any:
...
-        return args[0]
+@builder
+def build_settings(app: app.App) -> Settings:
+    return Settings(app.settings)
 
-    @classmethod
-    def build(cls, arg_name: ArgName, settings: Settings):
-        return settings.get(arg_name)
+
+Setting = object
```
tomchristie: Does `Setting = typing.Any` work?

tests/test_apischema.py
```diff
@@ -90,7 +90,7 @@ def set_complete(ident: int, complete: bool):  # pragma: nocover
 
 
 def test_api_schema():
-    schema = APISchema.build(app)
+    schema = core.get_builder(APISchema)(app)
```
tomchristie: Just build_apischema directly would be good.

apistar/settings.py
maackle: Nope, because later on this gets passed through `issubclass`, which expects a Class, not a type.

tomchristie:
Obsoleted by #235 - We won't have `build` methods at all once this work comes in.

PR #174, opened 2017-05-23
Fix console scripts entrypoint in setup.py

vickumar1981:


audiolion:
this was changed in https://github.com/tomchristie/apistar/pull/171 wonder if we should be changing scripts to match `cli` instead?

PR #177, opened 2017-05-23
Add missing type annotations

aqt01:
Add missing type annotations in `test_custom_command_with_int_arguments`

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -42,7 +42,7 @@ def test_custom_command():
     assert result.exit_code == 0
 
 
-def test_custom_command_with_int_arguments(a: int, b: int) -> int:
+def test_custom_command_with_int_arguments(*args: int) -> int:
     def add(a: int, b: int) -> int:
         click.echo(str(a + b))
         return (a+b)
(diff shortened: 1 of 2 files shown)


tests/test_commands.py
```diff
@@ -43,8 +43,9 @@ def custom(var):
 
 
 def test_custom_command_with_int_arguments():
-    def add(a: int, b: int):
+    def add(a: int, b: int) -> int:
         click.echo(str(a + b))
+        return (a+b)
```
tomchristie: We don't need to return a value here. I reckon just use `-> None` and don't return anything.

tests/test_http.py
```diff
@@ -284,8 +284,9 @@ def text_response():
     return '<html><h1>Hello, world</h1></html>'
 
 
-def data_response():
-    return {'hello': 'world'}
+def data_response() -> http.Response:
```
tomchristie: Let's not change the return type, since that breaks the intent of this test.
Probably need to just leave this one alone at the moment, as I think we need to do more work to properly deal with annotated data response types.

PR #179, opened 2017-05-24
Interactive title and description

brjadams:
Inject settings into APISchema, add title and description from settings object into Interactive API templates.

### DIFF
--- a/apistar/apischema.py
+++ b/apistar/apischema.py
@@ -20,7 +20,7 @@ from apistar.templating import Templates
 
 class APISchema(Document):
     @classmethod
-    def build(cls, app: App, settings: Settings={}, base_url: http.URL=None):
+    def build(cls, app: App, settings: Settings=Settings({}), base_url: http.URL=None):
         routes = app.routes
         url = get_schema_url(routes, base_url)
         content = get_schema_content(routes)
--- a/apistar/apischema.py
+++ b/apistar/apischema.py
@@ -20,13 +20,14 @@ from apistar.templating import Templates
 
 class APISchema(Document):
     @classmethod
-    def build(cls, app: App, settings: Settings=Settings({}), base_url: http.URL=None):
+    def build(cls, app: App, base_url: http.URL=None):
         routes = app.routes
         url = get_schema_url(routes, base_url)
         content = get_schema_content(routes)
-        schema = settings.get("SCHEMA", {})
-        title = schema.get("TITLE", None)
-        description = schema.get("DESCRIPTION", None)
+        settings = Settings(app.settings or {})
+        schema = settings.get('SCHEMA', {})
+        title = schema.get('TITLE', None)
+        description = schema.get('DESCRIPTION', None)
         return cls(url=url, content=content, title=title, description=description)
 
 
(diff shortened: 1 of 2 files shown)


brjadams:
Need to add default settings/schema to test.

apistar/apischema.py
```diff
@@ -14,16 +14,20 @@
...
     Route, RoutesConfig, primitive_types, schema_types, walk
 )
+from apistar.settings import Settings
 from apistar.templating import Templates
 
 
 class APISchema(Document):
     @classmethod
-    def build(cls, app: App, base_url: http.URL=None):
+    def build(cls, app: App, settings: Settings=Settings({}), base_url: http.URL=None):
```
tomchristie: In preference to this I'd suggest default to `None` instead, and do a...

    if settings is None:
        settings = Settings({})

apistar/templates/docs/layout/content.html
```diff
@@ -1,9 +1,7 @@
 <div class="row intro">
 <div class="col-md-6 intro-title">
     <h1>{{ document.title|default('API Star', True) }}</h1>
-    {% if document.description %}
-    <p>{{ document.description }}</p>
-    {% endif %}
+    <p>{{ document.description|default('description', True) }}</p>
```
tomchristie: Probably wrap `{% if document.description %}` around this `<p>` block - we don't want to provide a default text string here.

apistar/apischema.py
```diff
@@ -14,16 +14,20 @@
...
         routes = app.routes
         url = get_schema_url(routes, base_url)
         content = get_schema_content(routes)
-        return cls(url=url, content=content)
+        schema = settings.get("SCHEMA", {})
+        title = schema.get("TITLE", None)
+        description = schema.get("DESCRIPTION", None)
```
tomchristie: Probably prefer single quotes where possible (no particular reason, just prefer that stylistically)

brjadams: understood.

apistar/templates/docs/layout/content.html
brjadams: I believe it was this way previously. Will clean this up

ssteinerx:
Does this also address/close [Integrating `schema.description` with docs?](#211)?

tomchristie:
@ssteinerx I'm not sure that you meant to link back to this same ticket?

ssteinerx:
Whoops, I was trying to link from the pull request about it and got the wrong tab!

tomchristie:
That ticket is regarding field descriptions, rather than the top-level overall description.
Related, but not the same, no. :)

ssteinerx:
@tomchristie Ah, thanks, sorry about that, just re-read the code in this pull request.  Just started using this yesterday and have only poked around enough to realize that my descriptions weren't coming out in the /docs.

tomchristie:
I think this is a valid approach yup. If it gets refactored for 0.2, and gets appropriate docs, then I don't see why it shouldn't go in. Thanks!

tomchristie:
Closed via https://github.com/encode/apistar/commit/4c95797a2c13d0b379c8f4027c0686d8f4442532

Does still need documenting tho.

PR #182, opened 2017-05-24
Skip configuration of Django if already configured

apollo13:
If Django is already configured by other means there should be no need to reconfigure it.

### DIFF
--- a/apistar/backends/django_backend.py
+++ b/apistar/backends/django_backend.py
@@ -13,7 +13,7 @@ class DjangoBackend(object):
 
     @classmethod
     def build(cls, settings: Settings):
-        if not cls._loaded:
+        if not cls._loaded and not django_settings.configured:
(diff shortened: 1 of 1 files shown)


apollo13:
Actually `django.setup()` looks idempotent enough that this addition should not have any effect. That said, there is no point in even trying to configure it if Django is already configured -- so one might even wanna raise an error… Feel free to close or merge, I do not have any strong feelings :D

tomchristie:
On balance I reckon we're probably better off with the current state, rather than quietly not-configuring. We can review if we bump into unexpected issues.

apollo13:
Damn, `settings.configure()` raises an error if already configured. would it be possible to configure it only if `INSTALLED_APPS` is in settings? Realistically speaking you will not have any models without configuring `INSTALLED_APPS` so that looks like an okayish sentinel.

Issue #183, opened 2017-05-24
[docs] Is this missing a return?


audiolion:
```python
from apistar import wsgi

def hello_world() -> wsgi.WSGIResponse:
    wsgi.WSGIResponse(
        '200 OK',
        [('Content-Type', 'text/plain')],
        [b'Hello, world!']
    )
```

Shouldn't there be a `return` from `hello_world()` fn of the `WSGIResponse`?

PR #184, opened 2017-05-24
[Issue-183]

danbgray:
There should be a return in the code in the readme for wsgi.WSGIResponse(..)


audiolion:
Fixes #183

tomchristie:
Thanks!

PR #185, opened 2017-05-25
Add include example to README

nirgn975:
I feel like `Include` are really useful and we need an example for them in the README file (even just to inform)


tomchristie:
Yas please!

PR #188, opened 2017-05-27
Fix #187: ModuleNotFoundError: No module named 'django' when using SQLAlchemy

jriggins:
Fix for #187.  I'm sure that there is a more elegant way to fix this, but it seems to have gotten the job done.

PR #189, opened 2017-05-27
Add compat entry for urllib3 due to new requests version

audiolion:
Requests 2.16.0 removes vendored packages, this compat entry allows for older and newer versions of requests. Can be removed once we pin requests at `2.16.0`

Fixes errors in #188 PR and is pretty much the same fix used in my [rest framework PR](https://github.com/encode/django-rest-framework/pull/5185)


audiolion:
The change was reverted in 2.16.1

PR #192, opened 2017-06-03
Fix incorrect error text

ewjoachim:
`-f` does not exist but `--force` does.


ewjoachim:
Build seems broken because of a blank line missing in schema.py, fixing in #193

tomchristie:
Good catch, thanks!

tomchristie:
Thanks @ewjoachim - re-ran the build, all clean now.

PR #195, opened 2017-06-07
Adds support for callable default values in Schema definitions

dnmellen:
I think sometimes is very handy to be able to use functions as default values when you are defining a schema:

    def generate_uuid():
        return str(uuid.uuid4())

    class Artist(schema.Object):
        properties = {
            'id': schema.String(max_length=100, default=generate_uuid),
            'name': schema.String(max_length=100),
        }

This way I can use the schema as a template for creating new objects. I don't know if this is beyond the scope of the schemas but It would be nice to have this feature.

Thanks in advance!

### DIFF
--- a/apistar/schema.py
+++ b/apistar/schema.py
@@ -240,7 +240,10 @@ class Object(dict):
             except KeyError:
                 if hasattr(child_schema, 'default'):
                     # If a key is missing but has a default, then use that.
-                    self[key] = child_schema.default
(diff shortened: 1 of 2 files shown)


tomchristie:
Thanks so much for stepping up!
At the moment I'm going to call this out of scope. It's problematic in particular because we'd end up eg. indicating a fixed default value in the api documentation which wouldn't correspond to the actual behavior.

PR #201, opened 2017-06-08
Use all the Django settings

globophobe:
I needed to change the AUTH_USER_MODEL setting. Maybe it's useful to use all the Django settings.

### DIFF
--- a/apistar/backends/django_backend.py
+++ b/apistar/backends/django_backend.py
@@ -14,8 +14,7 @@ class DjangoBackend(object):
     @classmethod
     def build(cls, settings: Settings):
         if not cls._loaded:
-            django_settings.configure(INSTALLED_APPS=settings.get('INSTALLED_APPS'),
-                                      DATABASES=settings.get('DATABASES'))
+            django_settings.configure(**settings)
             setup()
             cls._loaded = True
         dj = cls()


ssteinerx:
Low impact, code-wise, but I'm writing a Django API Star ✨🚀✨🌟  demo app, and needed this, too.

tomchristie:
Good call, yes.
I've got a better picture now of how API Star will fit into Django land, so expect some adaptation in the future, but this is a worthwhile change for now.
Thanks!

  😄

tomchristie:
I'm gonna roll this back for 0.2, and just explicitly include `AUTH_USER_MODEL`.
We should limit the settings to anything that Django core actually uses to affect models.

globophobe:
I'm looking forward to what you've got in the pipeline for Django.

tomchristie:
0.2 does the groundwork of allowing us to support multiple different frameworks, tho currently that extends to asyncio and wsgi options.

Full "run inside Django" and "run inside Flask" options are (ideally?) planned for a later release.

The Django ORM support *is* better in 0.2 tho, since we can now support context managed components, allowing us to introduce a `Session` that handles `commit`/`rollback` behaviors, and properly cleans up connections.

Issue #122, opened 2017-04-26
Issue after upgrading to 0.1.13


maziarz:
After upgrading this happend. Python 3.5.1

```
>>> from typing import Type
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'Type'
 ```

```bash
$ apistar run
Traceback (most recent call last):
  File "/usr/local/bin/apistar", line 7, in <module>
    from apistar.main import main
  File "/usr/local/lib/python3.5/site-packages/apistar/__init__.py", line 8, in <module>
    from apistar.app import App
  File "/usr/local/lib/python3.5/site-packages/apistar/app.py", line 7, in <module>
    from apistar import commands as cmd
  File "/usr/local/lib/python3.5/site-packages/apistar/commands/__init__.py", line 2, in <module>
    from apistar.commands.new import new
  File "/usr/local/lib/python3.5/site-packages/apistar/commands/new.py", line 8, in <module>
    from apistar import schema
  File "/usr/local/lib/python3.5/site-packages/apistar/schema.py", line 2, in <module>
    from typing import Any, Dict, List, Tuple, Type, Union  # noqa
ImportError: cannot import name 'Type'
```

maziarz:
Yay! Fixed in 0.1.14 :dancing_men:

tomchristie:
We do need to be careful tho - the tests locally and on travis didn't pick this up, don't want to reintroduce it at a later date! 😐

maziarz:
Which Python versions are Travis testing with currently?

PR #210, opened 2017-06-13
Fix for Python defaulting to using stdlib typing

FFX01:
Fixes issue #122.

Solution taken from #306

### DIFF
--- a/apistar/apischema.py
+++ b/apistar/apischema.py
@@ -1,6 +1,8 @@
 import base64
 import inspect
-from typing import Dict, Optional, Type, cast
+from typing import Dict, Optional, cast
+if False:
+    from typing import Type
 from urllib.parse import urljoin
 
 import coreschema


tomchristie:
Not keen on `if False` here. Could we use `typing.Type` string literals for the `Type` annotation instead, maybe?

FFX01:
@tomchristie 

I'll try it out and see if it works.

FFX01:
@tomchristie 

Unfortunately, using a string literal here doesn't seem to work.

The issue is that when `from typing import ...` is called on my system it is defaulting to the stdlib `typing` module which does not include a declaration for `Type`. This is happening inside of the virtualenv because I believe module imports are first searched under the standard library and then `site-packages`.

So, I'm trying to figure out a fix that doesn't involve adding a new path to the global `$PYTHONPATH` env variable.

FFX01:
This is only an issue under Python 3.5. However, Python 3.5 is the standard version on most *buntu systems, so I think this is a major issue.

PR #216, opened 2017-06-22
Typing on schema.py files

suripoori:
[#87](https://github.com/tomchristie/apistar/issues/87)


tomchristie:
Thanks for having taken a crack at this! Deferred it, since things were going to change in v0.2, which is out now. Very welcome to look into adding any missing type annotations there.

suripoori:
Sure, will take a look.

PR #217, opened 2017-06-29
Adds type annotations for apistar/docs.py functions.

Rdbaker:
Let me know what you think! I was looking at issue #87 as a good starter task and the file I chose seemed to be a good place to start


tomchristie:
Looks good! 👍

PR #224, opened 2017-07-03
Use set comprehensions.

thedrow:
Instead of converting a list comprehension to a set we can use set comprehensions directly.
It's cleaner (and slightly faster).


tomchristie:
Very nice, yup. Don't think I knew this was a thing.

audiolion:
The reason it is not seen often is because the dict comprehension takes precedence. You can't make an empty set with the same notation `my_set = {}` because in this case it is interpreted as a dict comp. As a result we often see the `my_set = set()` syntax. But, when you don't include a key, value pair the interpreter can infer it is a set.

PR #226, opened 2017-07-06
added RequestData documentation

marcosfede:
added minimal http.RequestData information on README since I couldn't find how to parse a POST request data on it.


tomchristie:
👍  Thanks! Bit of an oversight there.
If you get a chance we should add the the fuller list of components towards the end of the README too.

PR #227, opened 2017-07-09
Turn off Werkzeug debugger with `run --no-debugger`

ryansb:
Some debugging workflows (like embedding epdb/pdb/etc) get tripped up
by the hooks the Werkzeug debugger uses to present the web UI debugger.
This PR adds a `--no-debugger` option to turn off Werkzeug but leave on the
auto-reload features.

To test this out, run any API* project with `apistar run --no-debugger` and do something that triggers an exception. The traceback will be logged, and a 500 response will be sent back, but the web debugger won't start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -935,6 +935,12 @@ both performance and for productivity.
 A development server is available, using the `run` command:
 
     $ apistar run
+    # Specify the port or interface via --port and --host
+    # Serve on port 9001 and use IPv6 only
+    $ apistar run --port 9001 --host ::1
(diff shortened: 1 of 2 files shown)


audiolion:
I guess there is no reference in the docs to providing the `--host` or `--port` options to the `apistar run` command, but there should probably be docs for them as well as docs that the `--no-debug` option exists, perhaps under the [development section](https://github.com/tomchristie/apistar#the-development-server)

tomchristie:
Fab, thanks! 👍

PR #229, opened 2017-07-14
Add support for schema.File

myarik:
I guess it could be helpful to have a file validation schema

### DIFF
--- a/apistar/schema.py
+++ b/apistar/schema.py
@@ -1,7 +1,11 @@
+import os
 import re
-from typing import Any, Dict, List, Optional, Tuple, Union, overload  # noqa
+from typing import Any, Dict, List, Optional, Union  # noqa
+
+from werkzeug.datastructures import FileStorage
 
 from apistar.exceptions import SchemaError, ValidationError
+
 
 # TODO: Error on unknown attributes
 # TODO: allow_blank?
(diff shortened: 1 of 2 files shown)


tests/schema/test_file.py
```diff
@@ -0,0 +1,55 @@
...
+
+def test_read_data():
+    data = 'some buffer'
+    response = client.post('http://example.com/upload/', files={'file': io.StringIO(data)})
+    assert response.status_code == 200
+    assert response.json() == {'body': 'some buffer'}
+
+
+def test_not_file():
+    response = client.post('http://example.com/upload/', data={'file': 'some buffer'})
```
audiolion: What would happen if you did a test like
```python
response = client.post('http://example.com/upload/', files={'file': 'some buffer'})
```
instead? Would that fail?

myarik: No, this request will return the 200 response because the method `parse_form_data` returns the FileStorage object with stream `some buffer`

tomchristie:
Thanks for submitting this!
It'll need some thinking about - I don't have the bandwidth to concentrate on this particular aspect right now, but I think it's something that'll come up a little further along the line.

apistar/schema.py
```diff
@@ -320,3 +324,44 @@ def __init__(self, value):
 
         if errors:
             raise SchemaError(errors)
+
+
+class File(object):
```
auvipy: no need to inherit from object in py3

tomchristie:
Obsoleted by #400 (We're finally back up and running again)

PR #231, opened 2017-07-20
Make both database backends optional

thedrow:
Right now importing one of the backends requires installing both.
This PR should fix that.

### DIFF
--- a/apistar/backends/__init__.py
+++ b/apistar/backends/__init__.py
@@ -1,14 +1,14 @@
 importable_backends = []
 
 try:
-    from apistar.backends.django_backend import DjangoBackend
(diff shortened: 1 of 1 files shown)


tomchristie:
Thanks! Needs a few exclusions to make the linting pass.

tomchristie:
Final thing is a codecoverage exclusion against that `__init__.py` file.
Thanks for your work here!

thedrow:
You don't really have a configuration file for coverage.
Do I edit https://github.com/tomchristie/apistar/blob/master/scripts/test#L17 instead?

tomchristie:
Obsoleted by #235

PR #238, opened 2017-08-09
Fixes #237

### LINKED ISSUE #237: mysql-python is not python3 compatible
In the readme, mysql-python is suggested as a mysql driver. However, it's not ported to python 3 yet ([reference](https://pypi.python.org/pypi/MySQL-python/1.2.5)). 

I installed `pymysql` instead, changed the prefix to `mysql+pymysql://...` and it worked well. At least I could create the table given in example. I guess the readme could be updated with this suggestion ?

brumar:
python-mysql is not ported to python 3 yet and pymysql seems to work as a replacement.


tomchristie:
👍  Thanks!

PR #245, opened 2017-08-19
Add inline Array type declaration

drpoggi:


tomchristie:
Yup! I wasn't sure if we should include the composite types, but let's roll with this. Thanks!

PR #248, opened 2017-08-20
Don't assign to prefix while iterating

### LINKED ISSUE #247: flatten_routes() is order-sensitive and gets the wrong path when Include precedes Route
When a list of routes that contains an `Includes` object before a `Route` object is passed to `apistar.core.flatten_routes`, the path of the Includes object is assigned to the Route object. For example:

```python
import apistar
from apistar import Include, Route

routes = [
    Include('/nested', [Route('/a', 'GET', lambda: 'a', name='a')]),
    Route('/x', 'GET', lambda: 'x', name='x'),
]

flat_routes = apistar.core.flatten_routes(routes)
assert flat_routes[0].path == '/nested/a'

# This check fails; the path picked up the parent route of the preceding Include
assert flat_routes[1].path == '/x' 
```

The cause of the bug is an inplace assignment to `prefix_path` in the `Include` branch of `flatten_routes()` [here](https://github.com/encode/apistar/blame/master/apistar/core.py#L75); `prefix_path` is then reused by any subsequent `Route` objects.

jlowin:
The path_prefix and namespace_prefix variables were assigned to during
iteration, leading to unwanted modification of later routes.
Closes #247

### DIFF
--- a/apistar/core.py
+++ b/apistar/core.py
@@ -78,10 +78,13 @@ def flatten_routes(routes: typing.Sequence[typing.Union[Route, Include]],
             flattened_routes.append(route)
         elif isinstance(item, Include):
             path, routes, namespace = item
-            path_prefix += path
+            includes_path_prefix = path_prefix + path
             if namespace is not None:
-                namespace_prefix += namespace + ':'
-            for route in flatten_routes(routes, path_prefix, namespace_prefix):
+                includes_namespace_prefix = namespace_prefix + namespace + ':'
+            else:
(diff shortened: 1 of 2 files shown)


jlowin:
@tomchristie got hung up on some lint errors but I think they're fixed now! 😃

tomchristie:
Looks great! I'll review more thoroughly once I'm at a desk. Fab stuff, thank you!

tomchristie:
👍  Awesome, thanks.

PR #250, opened 2017-08-22
Proposal: Add InjectionCaller to allow nested injection scopes

jlowin:
For better or worse, I often call API functions from elsewhere in my API in a nested fashion (API function A calls API function B). Dependency injection means that the "parent" function is handed any dynamic variables/parameters it needs, but the "child" function must have them supplied explicitly.

A concrete example of this is a GraphQL endpoint. The GQL endpoint is one of a number of HTTP-routed API functions that receive normal dependency injection. But the GraphQL endpoint might expose a number of mutations that themselves call (or are) other API functions. The endpoint, which receives the injected variables, doesn't directly call the mutations (graphene has its own resolver system), nor can it optimistically request all the dependencies that the nested API functions might possibly require. What we want in this case is a way to load the current "injection context" at runtime and reuse it to call other API functions.

The challenge of designing a component that allows this runtime injection is that we want the `state` of the "parent" call to be available to the "child" injection function, and components are typically bound/introspected before that state is known. For example, if the parent parsed an HTTP request header, we want that available to the child.

So here I have one proposal for how to make this work: the `InjectionCaller` component. I'm still getting my bearings with the framework so I'm not sure this is compliant with "the API Star way" and would welcome thoughts.

Here's an example, see the unit test for a more complete one:

```python

def api_function_1(user: AuthenticatedUser):
    """
    This is one of my api functions that uses injection to authenticate a user
    """
    do_something_with_authenticated_user(user)

def api_function_2(call_with_injection: InjectionCaller):
    """
    This is another api function that one way or another needs to call the first function
    but doesn't necessarily know what injected parameters it requires
    """
    # this call will properly parse the authenticated user even though it 
    # wasn't passed to this scope
    call_with_injection(api_fuction_1)
```

### DIFF
--- a/apistar/components/dependency.py
+++ b/apistar/components/dependency.py
@@ -51,6 +51,9 @@ class DependencyInjector(Injector):
         if resolvers is None:
             resolvers = []  # pragma: nocover
 
+        # make this injector available to components
+        initial_state[Injector] = self
+
         self.components = components
         self.initial_state = initial_state
         self.required_state = required_state
@@ -92,6 +95,9 @@ class DependencyInjector(Injector):
 
         # Combine any preconfigured initial state with any explicit per-call state.
         state = {**self._setup_state, **state}
+
+        # recursively store state
+        state['__injector_state__'] = state
 
         ret = None
         with ExitStack() as stack:
@@ -152,6 +158,13 @@ class DependencyInjector(Injector):
             # don't run any function. The value must be passed explicitly
             # in the `state` dictionary when calling `run()`.
             key = self.required_state[annotation]
+            func = None
+            return (key, func)
+
+        elif annotation is InjectorState:
+            # InjectionState requires special handling since its value depends
+            # on the context of this call
+            key = '__injector_state__'
             func = None
             return (key, func)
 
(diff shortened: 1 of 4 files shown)


apistar/components/dependency.py
```diff
@@ -450,3 +467,26 @@ def container_type(self,
...
+class InjectionCaller:
+
+    def __init__(self, injector: Injector, state: InjectorState) -> None:
+        self.injector = injector
+        self.state = state
+        self.is_async = isinstance(injector, AsyncDependencyInjector)
+
+    def __call__(self, func):
+        if asyncio.iscoroutinefunction(func):
+            return self.run_async(func)
```
tomchristie: This doesn't quite work out, since we actually want to use `await` on this, and call it from `async` functions only.

tomchristie:
Yup, we absolutely will need something along these lines.

For cases where you're explicitly calling into a function you can just include the requirements in both places, however there's cases where you want to be able to provide a generic "call into this *thing*" where we need this.

Not sure if this is how I'd want to tackle it or not - would need some further through before progressing it. In the meantime you could maintain an alternative Injector in your own codebase, and use that with the app, using `INJECTOR_CLS = ...`

jlowin:
Figured this was an early stab -- glad to hear it's on the roadmap! I'll close this for now and use this class for my own INJECTOR_CLS in the meantime, as you suggested. Thanks!

tomchristie:
👍

PR #251, opened 2017-08-22
Add check to see if parent_param is defined before accessing it

### LINKED ISSUE #246: Using `ParamName` in component can crash App
Hey @tomchristie, love this framework and am quickly trying to get up to speed. I've discovered that there is an implicit assumption that `ParamName` annotations are always used in conjunction with a `parent_param`, but it's easy to violate that assumption. 

Here is a contrived but illustrative example:

```python
import base64
import typing

import apistar
from apistar import Component, Route
from apistar.frameworks.wsgi import WSGIApp as App

values = {
    'a': 'foo',
    'b': 'bar',
}


def get_value(name: apistar.types.ParamName):
    return values.get(name)


GetValue = typing.NewType('GetValue', str)


def print_value(a: GetValue, b: GetValue):
    """
    Contrived function that retrievs a value based on the parameter name
    """
    return 'The value of a is {} and the value of b is {}'.format(a, b)


app = App(
    routes=[Route('/', 'GET', print_value)],
    components=[Component(GetValue, init=get_value)])

if __name__ == '__main__':
    app.main()
```
This results in:
```
❯ apistar run
 * Running on http://127.0.0.1:8080/ (Press CTRL+C to quit)
 * Restarting with stat
 * Debugger is active!
 * Debugger PIN: 211-517-262
127.0.0.1 - - [20/Aug/2017 09:18:45] "GET / HTTP/1.1" 200 -
 * Detected change in '/users/jlowin/miniconda3/lib/python3.6/site-packages/apistar/components/dependency.py', reloading
 * Restarting with stat
Traceback (most recent call last):
  File "/users/jlowin/miniconda3/lib/python3.6/site-packages/apistar/components/dependency.py", line 88, in run
    steps = self._steps_cache[func]
KeyError: <function get_value at 0x10ca09e18>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/jlowin/miniconda3/bin/apistar", line 11, in <module>
    sys.exit(main())
  File "/users/jlowin/miniconda3/lib/python3.6/site-packages/apistar/main.py", line 34, in main
    app = load_app()
  File "/users/jlowin/miniconda3/lib/python3.6/site-packages/apistar/main.py", line 17, in load_app
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 205, in _call_with_frames_removed
  File "app.py", line 31, in <module>
    components=[Component(GetValue, init=get_value)]
  File "/users/jlowin/miniconda3/lib/python3.6/site-packages/apistar/frameworks/wsgi.py", line 62, in __init__
    super().__init__(**kwargs)
  File "/users/jlowin/miniconda3/lib/python3.6/site-packages/apistar/frameworks/cli.py", line 49, in __init__
    initial_state=initial_state
  File "/users/jlowin/miniconda3/lib/python3.6/site-packages/apistar/frameworks/cli.py", line 91, in preload_components
    component = injector.run(func)
  File "/users/jlowin/miniconda3/lib/python3.6/site-packages/apistar/components/dependency.py", line 90, in run
    steps = self._create_steps(func)
  File "/users/jlowin/miniconda3/lib/python3.6/site-packages/apistar/components/dependency.py", line 197, in _create_steps
    input_values[param.name] = parent_param.name
AttributeError: 'NoneType' object has no attribute 'name'
```

I'm not 100% clear on the internals yet but I think the solution is to simply change [this line](https://github.com/encode/apistar/blob/1a2519c66aa1bd9a304e040a60d02e644d045fcc/apistar/components/dependency.py#L196) from 
```python
if param.annotation is ParamName:
```
to 
```python
if param.annotation is ParamName and parent_param:
```
If so, I'm happy to submit a PR to that effect (the following if statement would also need to be adjusted).

In case you're wondering why this came up: I wanted to use `ParamName` directives to pull information out of a `dict` describing an authenticated user, in exactly the same way that API Star retrieves `Headers` based on the param name (like `authorization`).

jlowin:
Simple fix to close #246 (with test)

### DIFF
--- a/apistar/components/dependency.py
+++ b/apistar/components/dependency.py
@@ -193,11 +193,11 @@ class DependencyInjector(Injector):
 
         # Add the steps required to satisfy each parameter in the function.
         for param in inspect.signature(func).parameters.values():
-            if param.annotation is ParamName:
+            if param.annotation is ParamName and parent_param:
                 input_values[param.name] = parent_param.name
                 continue
 
-            if param.annotation is ParamAnnotation:
+            if param.annotation is ParamAnnotation and parent_param:
                 input_values[param.name] = parent_param.annotation
                 continue
 
--- a/tests/test_dependency.py
+++ b/tests/test_dependency.py
@@ -1,8 +1,10 @@
-from typing import List
+from typing import List, NewType
 
 import pytest
 
-from apistar import Component, Route, TestClient, exceptions, http, typesystem
+from apistar import (
+    Component, Route, TestClient, exceptions, http, types, typesystem
+)
 from apistar.frameworks.asyncio import ASyncIOApp
 from apistar.frameworks.wsgi import WSGIApp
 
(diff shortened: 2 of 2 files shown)


tomchristie:
Obsoleted by the #400 redesign.

Issue #78, opened 2017-04-20
Auth & Permissions


tomchristie:
Right now authentication and permissions have to be handled as custom components or inside the view handler. We'll want to provide some out of the box options here, such as JWT.

maziarz:
+1 for [oauthlib](https://github.com/idan/oauthlib) integration

piotr-szpetkowski:
I'd be interested in [python-jose](https://github.com/mpdavis/python-jose)

tomchristie:
Good pointers, thanks folks!

maziarz:
I am very curious about the Permissions part as i have been experiencing quite many different approaches from the JS world, where ACL's are very common and combined with Role assignments.

tomchristie:
@maziarz Don't expect us to present anything super-abstracted at this point. Decisions about 'roles' and permission levels are likely to be left in application land - we'll just be making sure that there's a good general purpose answer to applying whatever kind of policy you do want either on a global or a per-view level.

maziarz:
I nearly forgot sessions..

audiolion:
Was this addressed in the 0.2.0 release? I am hitting this point right now where I will be calling the same function at the start of every view function to check a JWT for authorization and handling before proceeding, I can build my own JWT component but I wanted to ask here first if maybe its something that I should just be helping build into the library with a PR.

Do you know what you are looking for in a JWT handler implementation? I am assuming a component that uses the `pyjwt` library to encode and decode, basing its secret on an apistar setting the user sets and throwing an exception on import if the user hasn't defined the setting.

tomchristie:
Closing this off given that 0.6 is moving to a framework-agnostic suite of API tools, and will no longer include the server. See https://discuss.apistar.org/t/api-star-as-a-framework-independant-tool/614 and #624.

PR #255, opened 2017-08-25
(feat) JWT Authorization Component

audiolion:
Hi Tom!

As a follow up from #78 I implemented a JWT Component as part of a new file under the components module called "authorization.py" (to hold common auth components, OAUTH 1.0, OAUTH 2.0, Basic Auth, Digest Auth, AWS Signature, .. sky's the limit).

I tried to follow your current architecture patterns as best I could.

I wasn't sure if I should be including messages about why the request failed when injecting the JWT token / decoding it.

This is just a first pass and doesn't include the other half (creating a new JWT from a payload). Not hard to add but I wanted to make sure I was on the right track. Also wasn't sure how to name it, you have two components, the DecodedJWT and the EncodedJWT, maybe I should change them to that, but, `my_func(request: http.Request, token: authorization.DecodedJWT)` seems worse than just `authorization.JWT`.

The other question I had was about returning the Component, I put the payload inside the Component as `payload`, which requires `token.payload['key']` to get at the data, there is no other information with the JWT (I don't know why we would include the secret or the algorithm used) so I wasn't sure if I should just return a `typing.Dict[str, typing.Any]` "json" dict, but I felt that broke the idea that we are injecting a component called JWT  but is in fact just an anonymous object.

One last thing, I didn't see any use for adding an `Authorization` interface to `interfaces.py`, there were no methods on the JWT class that I would use. But now that I think of it, I could add a `decode` and `encode` method and refactor the JWT component to handle both cases, the only issue being how to call two different init methods on the same component.

edit: Latest commit included DecodedJWT/EncodedJWT Components, one is an http method to be injected and the other is a component to be imported and used. Only hangup I have with the imported component is that mypy will fail unless the person passing the payload argument to EncodedJWT wraps it in the `JSON` type. Not sure if that is what we want or not. My latest commit removes it.

### DIFF
--- a/apistar/components/authorization.py
+++ b/apistar/components/authorization.py
@@ -1,0 +1,38 @@
+import typing
+
+import jwt
+
+from apistar import exceptions, http
+from apistar.types import Settings
+
+Token = typing.NewType('Token', str)
+
+
+class JWT():
+    DEFAULT_SETTINGS = {
+        'JWT_ALGORITHMS': ['HS256'],
+    }
+
+    def __init__(self,
+                 token: Token,
+                 settings: Settings) -> None:
+        authorization_settings = settings.get('AUTHORIZATION', self.DEFAULT_SETTINGS)
+        secret = authorization_settings.get('JWT_SECRET', None)
+        if secret is None:
+            msg = 'The JWT_SECRET setting under AUTHORIZATION settings must be defined.'
+            raise exceptions.ConfigurationError(msg) from None
+        algorithms = authorization_settings.get('JWT_ALGORITHMS', ['HS256'])
+        try:
+            payload = jwt.decode(token, secret, algorithms=algorithms)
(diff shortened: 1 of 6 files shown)


tomchristie:
Good stuff. We're at the point where we ought to start pushing towards third party packages rather functionality within core. I expect the Django ORM and SQLAlchemy backends will get pushed out to third party pckages two.

We're also currently missing a generic authentication API similar to what we have in REST framework. I should be addressing this fairly shortly.

audiolion:
ok so break this out into a third-party package?

tomchristie:
> ok so break this out into a third-party package?

That'll certainly be one aspect, yup.

Secondly, be prepared to do a bit of rejigging once a proper auth interface lands.

audiolion:
haha alright 👍 ill close this then

PR #256, opened 2017-08-25
Add unit tests for annotated QueryParams and also support for parsing function args from request data

jlowin:
The README says:

> Parameters which do not correspond to a URL path parameter will be treated as
> query parameters for `GET` and `DELETE` requests, or part of the request body
> for `POST`, `PUT`, and `PATCH` requests.

However, there are no unit tests for this behavior, and indeed as far as I can tell it is only currently implemented for query parameters, not the request data dictionary.

This PR does two things:
1. Adds unit tests for automatically parsing scalar-annotated and non-annotated function arguments from query parameters
2. Adds support (and unit tests) for automatically parsing scalar-annotated and non-annotated function arguments from the request data dictionary. In the current code, function arguments are matched first against the URL keywords, then the query parameters. This extends that to check a third place: the request data object.

### DIFF
--- a/apistar/components/dependency.py
+++ b/apistar/components/dependency.py
@@ -367,38 +367,47 @@ class HTTPResolver(Resolver):
     def empty(self,
               name: ParamName,
               kwargs: KeywordArgs,
-              query_params: http.QueryParams) -> str:
+              query_params: http.QueryParams,
+              data: http.RequestData) -> str:
         """
         Handles unannotated parameters for HTTP requests.
-        These types use either a matched URL keyword argument, or else
-        a query parameter.
+        First the URL keywords are checked for a match, then the query
+        parameters, and finally the request data (if it is a dictionary).
 
         Args:
             name: The name of the parameter.
             kwargs: The URL keyword arguments, as returned by the router.
             query_params: The query parameters of the incoming HTTP request.
+            data: The incoming HTTP request's data payload.
 
         Returns:
             The value that should be used for the handler function.
         """
         if name in kwargs:
             return kwargs[name]
-        return query_params.get(name)
+        elif name in query_params:
+            return query_params[name]
+        elif isinstance(data, dict):
+            return data.get(name)
+        else:
+            return None
 
     def scalar_type(self,
                     name: ParamName,
                     kwargs: KeywordArgs,
                     query_params: http.QueryParams,
+                    data: http.RequestData,
                     coerce: ParamAnnotation) -> typing.Any:
         """
         Handles `str`, `int`, `float`, or `bool` annotations for HTTP requests.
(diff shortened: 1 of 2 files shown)


jlowin:
If a developer is concerned with the situation you describe, then APIStar already provides the tools to enforce how and where arguments are parsed by annotating them as either `QueryParam` or `RequestData` (or a `ParamName` derivative thereof). This PR only affects the case where the developer has specifically left a function argument unannotated.

In your specific example, the query param would be honored over the data, even before this PR. The search hierarchy is: URL keywords, then query params, then data. The last bit - checking data - is all this PR modifies.

audiolion:
@jlowin yeah nothing wrong with the implementation and see the logic and how it follows. I was just thinking of the situation where the developer isn't aware of the precedence order it could cause an issue down the road with someone.

tomchristie:
Actually the docs are wrong, not the code. In moving to 0.2 I arrived at the conclusion that it'd be more consistent to instead use:

* Scalar types default to query params
* Lists and Dicts default to request body

I've evidently failed to adjust the docs accordingly.

We will also need:

* Support for overriding the defaults.
* Erroring on invalid cases. (Attempting to use request body on GET requests.)

jlowin:
@tomchristie understood.

The motivation for this is that building an API for `POST` requests is more currently more cumbersome than for `GET` requests, since it requires either parsing keys from the `RequestData` object (which means your function signature and your API signature are not the same) or a custom `Component` similar to `QueryParam` that does that parsing for you. I've gone with the latter since it keeps the signatures in line but keep wishing I had easy type annotation and coercion the way `GET` request parameters do.

In other words this:
```python
def my_GET_function(x: int, y: str):
    ...
```
is much nicer to write than this, which also requires defining and registering the custom component.
```python
def my_POST_function(x: CustomDataParam, y: CustomDataParam):
    x = int(x)
    y = str(y)
   ...
```

I would argue in favor of having some mechanism for parsing the request data, similar to query params. A few thoughts for discussion:
1. we could enforce a strict split where empty-or-builtin-annotated `GET`/`DELETE` parameters come from query params (which is the current API Star behavior), and `PUT`/`POST`/`PATCH` come from data (which would be new behavior). This would also address @audiolion's concern about overlap.
2. we could create a decorator that would allow this behavior to be turned on explicitly, something like:

```python
@args_from_query_params
def my_function(x, y:str, z:http.QueryParams):
    assert x = z['x']
    assert y = z['y']

@args_from_request_data
def my_function(x, y:str, z:http.RequestData):
    assert x = z['x']
    assert y = z['y']
```

tomchristie:
Agreed. Likely we'll have some kind of annotation that allows you to specify explicitly where the parameter is taken from.

tomchristie:

PR #259, opened 2017-08-26
allow changing folder in which app.py resides

### LINKED ISSUE #258: run cli from different folder
Hi, I'd like to run the cli from a different folder as `app.py`

I tend to put all my code in a `src` folder and run all cli commands from the parent. usually I either `pip install . -e` the package in the `src` folder or set the `PYTHONPATH=src`. for example with django, I would just use the `django-admin` cli command with the `DJANGO_SETTINGS_MODULE` env var set to `mypackage.settings` where `mypackage` is in the `src` folder

some potential options:
- `APISTAR_FOLDER` env var (or some other name). this aligns with django and also flask cli which has a `FLASK_APP` env var. for example: 
    - `export FLASK_APP=hello.py; export FLASK_DEBUG=1; flask run`
- a cli parameter. docker-compose for example has a `--project-directory` to specify an alternate working directory

I'd like the env var solution better.

I'd be happy to contribute this feature if you let me know which option you'd prefer @tomchristie

AlJohri:
This PR allows changing the folder in which app.py resides by setting `APISTAR_FOLDER` environment variable. If you wanted to allow changing the filename of `app.py` as well, you could supply the full path to the file and rename the environment variable to `APISTAR_APP`, more akin to flask.

fixes #258

### DIFF
--- a/apistar/main.py
+++ b/apistar/main.py
@@ -10,9 +10,10 @@ from apistar.frameworks.cli import CliApp
 from apistar.interfaces import App
 
 
-def load_app(folder) -> App:
+def load_app(filepath) -> App:
+    folder = os.path.dirname(filepath)
     sys.path.insert(0, folder)
-    spec = importlib.util.spec_from_file_location("app", "%s/app.py" % folder)
+    spec = importlib.util.spec_from_file_location("app", filepath)
     module = importlib.util.module_from_spec(spec)
     spec.loader.exec_module(module)
     app = getattr(module, 'app', None)
@@ -30,9 +31,9 @@ def default_app() -> App:
 
 
 def main() -> None:  # pragma: nocover
-    folder = os.getenv('APISTAR_FOLDER', os.getcwd())
-    if os.path.exists('%s/app.py' % folder):
-        app = load_app(folder)
+    filepath = os.getenv('APISTAR_APP', os.path.join(os.getcwd(), 'app.py'))
+    if os.path.isfile(filepath):
+        app = load_app(filepath)
     else:
         app = default_app()
     app.main()
--- a/apistar/main.py
+++ b/apistar/main.py
@@ -10,9 +10,9 @@ from apistar.frameworks.cli import CliApp
 from apistar.interfaces import App
 
 
-def load_app() -> App:
-    sys.path.insert(0, os.getcwd())
-    spec = importlib.util.spec_from_file_location("app", "app.py")
+def load_app(folder) -> App:
+    sys.path.insert(0, folder)
+    spec = importlib.util.spec_from_file_location("app", "%s/app.py" % folder)
     module = importlib.util.module_from_spec(spec)
     spec.loader.exec_module(module)
     app = getattr(module, 'app', None)
@@ -30,8 +30,9 @@ def default_app() -> App:
 
 
 def main() -> None:  # pragma: nocover
-    if os.path.exists('app.py'):
-        app = load_app()
+    folder = os.getenv('APISTAR_FOLDER', os.getcwd())
+    if os.path.exists('%s/app.py' % folder):
+        app = load_app(folder)
     else:
         app = default_app()
     app.main()


tomchristie:
I've not given this enough thought to okay it yet, but either way APISTAR_APP would be preferable, yup. 😀

AlJohri:
sure, let me know. this issue is pretty important to me because I have so many files at the top level like Dockerfile, docker-compose.yml, docker-compose.override.yml, server.ini, requirements.txt, docs, envs, README.md, .python-version, .gitignore, and on and on and on...

AlJohri:
hi @tomchristie, bumping this issue

tomchristie:
I think I'd accept this if the env variable was `APISTAR_APP`, and the documentation noted it in the "Deployment > The Development Server" section.

In the meantime you can use the `if __name__ == '__main__': app.main()` pattern, and invoke the app.py directly with Python.

We might want to rethink the `apistar` cli more thoroughly in the future, but this will do for now.

rougeth:
This PR would make only `apistar run` works or `test` too? I'm asking this becouse the `if __name__ == '__main__': app.main()` approach doesn't works for running the tests.

AlJohri:
@rougeth I think it should affect tests too

rougeth:
Great @AlJohri!! :)

AlJohri:
Ah, that’s unfortunate. Hopefully we can make it depend on the same variable

rougeth:
@AlJohri any updates on this? Did you saw that @tomchristie asked to change the name of the variable from `APISTAR_FOLDER` to `APISTAR_APP`? This would be reaaally useful here, let me know if you need help.

AlJohri:
I'll check it out - I'm not that experienced with the codebase but we basically need to get the `APISTAR_APP` into the apistar.commands.test method, probably through DI of some sort. I can also just do another os.getenv but it would probably be best to set it at initialization time and use it throughout

rougeth:
@AlJohri I believe making `apistar test` aware of `APISTAR_APP` variable is another issue (if I understood correctly, you're talking about the tests, right?).  At my particular case, the way it is right now works. We could work on tests after that.

AlJohri:
@rougeth I think its sort of related since I want to set the value of APISTAR_APP at some initialization stage so it can be used throughout (including within tests), but to get things moving I've  added a small change to use `APISTAR_APP` instead of `APISTAR_FOLDER`.

I'll update the docs the next chance I get

JulienPalard:
Why not allowing to give a module to apistar, like `apistar run mypackage.app`? It would allow people to move `api.py` to a package, name it as they want, and to import the package from their tests?

pslacerda:
The workaround of using `if __name__ == '__main__': app.main()` requires the top level directory at the PYTHONPATH. Even if it is on the path, the workaround fails if relative imports to a parent package are used.

I agree to use some `APISTAR_APP` environment variable pointing to the `App` object, but it isn't really a variable, it is fixed. The `App` object remains the same across deployments, environment variables are supposed to hold values that can change possibly.

The problem that this issue address is to be able to run the app from anywhere. So I guess that is better to stick with the gunicorn approach:

    gunicorn webapp.somemodule:app \
        --worker-class=meinheld.gmeinheld.MeinheldWorker. 

Rendering something like:

    apistar run myapp.gateway.app:app

gunicorn uses [`import_app()`](https://github.com/benoitc/gunicorn/blob/6ba2ebe545f3540acb06cc768c74506e13c8901d/gunicorn/util.py#L342) (somewhat similar to the PR) to acquire the application object using the "package.module:object" notation.

#350 is related to this. This issue made impossible for me to use the `apistar run` command in debug mode. What can I do?

AlJohri:
hi all, sorry I don't have time to finish this PR as I haven't actually used apistar in any projects yet. will revisit this if I ever try apistar again. please feel free to open a new PR

pslacerda:
My workaround was simply

```python
from werkzeug.serving import run_simple
from mypackage.app import app

run_simple('0.0.0.0', 8080, app)
```

PR #262, opened 2017-08-27
Fixed reverse_url example in README.md

Lobosque:
* Updated the `reverse_url` example to reflect the current `reverse_url` signature, using a dict instead kwargs.
* Matched the route named argument to be the same as the `get_player_details` function argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,7 +298,7 @@ def get_all_players(router: Router):
     player_list = [
         {
             'name': player.name,
-            'url': router.reverse_url('get_player_details', player_name=player.name)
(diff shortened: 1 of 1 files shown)


Lobosque:
I also thought about adding the `get_score` and `get_players` function so that the reader can just copy / paste and see it working. What do you think @tomchristie?

tomchristie:
Great stuff, thanks!

> I also thought about adding the get_score and get_players function so that the reader can just copy / paste and see it working. What do you think @tomchristie?

Having the players as an in-memory dict might allow use to keep the example small enough to do this, yeah. We might not even need to make function calls if the sample code is minimal enough.

PR #267, opened 2017-08-31
Clean up README.md syntax highlighting

drpoggi:
👍


tomchristie:
Good call, yup. Thanks!

Issue #211, opened 2017-06-16
Integrating `schema.description` with docs?


GeoTob:
Is there a way to populate the `description` column for each field in the automatically generated docs:

![documentation](https://github.com/tomchristie/apistar/raw/master/docs/img/apistar.png)

I have tried to use a custom schema like below without any success:
```python
class Name(schema.String):
    description = 'Name for greeting'


def greet(name:Name):
    ...
```

tomchristie:
Yup, that's intended but hasn't made it's way in yet.

aidiss:
How can  I add it manually?

audiolion:
@aidiss fork the repo and add it in manually, or make a PR to this library to add support for it

ssteinerx:
Is the intention to support `description` via custom schema as shown by @GeoTob or some other method?  

Also, will e.g. `default` be able to be set and used in a similar manner?

tomchristie:
I'm going to close this off, since we don't have the docs back in yet, *but* worth noting that the redesigned type system *does* support `description=...` on the validators, which we'll use to pull this info in, once we've got the redesigned docs going.

PR #268, opened 2017-08-31
Display annotation description in documentation field

erik-sn:
I'm not sure if there is an existing idea but I made this change on a personal project and thought I would see if it was worth implementing. Appreciate any feedback you have.

For an app.py of: https://gist.github.com/erik-sn/7e9d50b4591418198f642364608b8f4e

Results in documentation :

image


erik-sn:
based on this issue: #211

tomchristie:
Yup - 100%.

We'll aim to get away from having separate "coreschema" instances at all in due course, but this is the right thing to do in the interim. Thanks!

PR #270, opened 2017-09-01
Add Optional type class

meadsteve:
Provides a class `Optional` that can wrap another type so Objects can have poperties that aren't required but that are type checked.
```python
class ObjectWithOptionalValues(typesystem.Object):
    properties = {
        'name': typesystem.string(),
        'some_optional_number': typesystem.Optional(typesystem.number())
    }
```

### DIFF
--- a/tests/typesystem/test_object.py
+++ b/tests/typesystem/test_object.py
@@ -25,6 +25,13 @@ class HighScore(typesystem.Object):
     }
 
 
+class ObjectWithOptionalValues(typesystem.Object):
+    properties = {
+        'name': typesystem.string(),
+        'some_optional_number': typesystem.Optional(typesystem.number())
+    }
+
+
 def basic_object(score: HighScore):
     return score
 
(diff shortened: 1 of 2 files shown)


kianxineki:
Would not a general configuration argument be better for all types?
```python
typesystem.number(required=False)
```

meadsteve:
@kianxineki the inspiration here was python's typesystem. So that it's like `Optional[str]`.

The other benefit is that the wrapped type doesn't need to be aware of the required property.

tests/typesystem/test_object.py
```diff
@@ -99,6 +106,23 @@ class test_object_default():
...
+
+
+class test_optional_keys_are_still_validated():
+    with pytest.raises(exceptions.TypeSystemError) as exc:
+        value = ObjectWithOptionalValues({
+            'name': 'steve',
+            'some_optional_number': 'hello'
+        })
+    assert str(exc.value) \
+        == '{\'some_optional_number\': \'Must be a valid number.\'}'
```
kianxineki: for clarity of code would not it be better this way? (79 columns)
```python
assert str(exc.value) == "{'some_optional_number': 'Must be a valid number.'}"
```

audiolion: +1 for all on the same line, the linter might error on double quotes though

meadsteve: works for me 👍

meadsteve:
Anything more I should do on this PR?

mammuth:
Recently a  [`required`](https://github.com/encode/apistar/commit/74498d629ee032796feae16564c6a441367452a3) list was added to `typesystem.object` which makes this implementation of optional/required fields obsolete, doesn't it? :confused:

PR #271, opened 2017-09-04
Improve schema command

bikeshedder:
- Fix JSON output of schema command
- Improve error handling for unsupported schema formats

### DIFF
--- a/apistar/commands/schema.py
+++ b/apistar/commands/schema.py
@@ -9,7 +9,7 @@ codecs = {
 
 
 def schema(schema: Schema,
-           format: str='corejson') -> bytes:
+           format: str='corejson') -> str:
     """
     Generate an API schema.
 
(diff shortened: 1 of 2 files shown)


tomchristie:
The improved error handling is clear to me, we should def. have that, yup!
Not obvious at first what exact behavioral change "fix JSON output" relates to?
There's a linter error and a couple of test failures that you'd need to look into.
Optionally you could submit these two separate issues in separate pull requests,
which'd make it easier for us to start to pull in at least some of the work.

bikeshedder:
The problem with returning a bytestring is that it outputs `b'...'` on the command line. This makes it impossible to use it to extract the schema like that:

    `$ apistar schema > api-schema.json`

I especially stumbled across this as I wanted to get pretty printing using `jq`:

   `$ apistar schema | jq`

I forgot to update the tests which should test for unicode strings.

bikeshedder:
I just fixed the linting errors and tests. You're right that I could split this into two PRs but as both touch the same file with two adjacent lines I wanted to keep it simpler to merge them.

tomchristie:
Fab, thanks!

Issue #120, opened 2017-04-26
Graceful handling of malformed RequestData


tomchristie:
Invalid JSON or invalid form content should result in a 400 response.

Also consider if we want to tweak the behavior around empty content at all (at the moment we return `None`)

piotr-szpetkowski:
👍 Right now if I try to e.g. send a POST request with no data I get 500:

```
apistar/http.py", line 187, in build
    value = data[arg_name]  # type: ignore
TypeError: 'NoneType' object is not subscriptable
```

pavelholvi:
Hey, has there been any updates on this? How do I reproduce this? Doing something like
```
>>> requests.post('http://127.0.0.1:8080/')
<Response [200]>
```
works OK.

tomchristie:
Eg. Make a POST request with a content-type header set to application/json, and a body containing some invalid bytes that cannot be parsed as JSON.

pavelholvi:
```
>>> requests.post('http://127.0.0.1:8080/',  json=b'123')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/isaevpd/py_virtual_envs/apistar_test/lib/python3.6/site-packages/requests/api.py", line 112, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/Users/isaevpd/py_virtual_envs/apistar_test/lib/python3.6/site-packages/requests/api.py", line 58, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Users/isaevpd/py_virtual_envs/apistar_test/lib/python3.6/site-packages/requests/sessions.py", line 494, in request
    prep = self.prepare_request(req)
  File "/Users/isaevpd/py_virtual_envs/apistar_test/lib/python3.6/site-packages/requests/sessions.py", line 437, in prepare_request
    hooks=merge_hooks(request.hooks, self.hooks),
  File "/Users/isaevpd/py_virtual_envs/apistar_test/lib/python3.6/site-packages/requests/models.py", line 308, in prepare
    self.prepare_body(data, files, json)
  File "/Users/isaevpd/py_virtual_envs/apistar_test/lib/python3.6/site-packages/requests/models.py", line 458, in prepare_body
    body = complexjson.dumps(json)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
[...]

PR #277, opened 2017-09-08
Graceful handling of malformed RequestData Issue #120

pavelholvi:
When sending a request with empty JSON data we get JSONDecodeError with `wsgi_client` and AttributeError with `async_client`.

In the first case body is parsed into empty string, `body.decode` works fine but `json.loads(decoded_body)` fails.
In the second case body is parsed into `None` and we end up with `AttributeError` on `body.decode()`.

Likewise, sending malformed data results in unhandled `JSONDecodeError` for both clients.

PR handles those cases to result in 400 response code and appropriate error message. Covered by tests.
https://github.com/encode/apistar/issues/120

### DIFF
--- a/apistar/components/umi.py
+++ b/apistar/components/umi.py
@@ -107,11 +107,11 @@ async def get_request_data(headers: http.Headers, message: UMIMessage, channels:
     elif mimetype == 'application/json':
         body = await get_body(message, channels)
         if not body:
-            raise exceptions.EmptyJSON()
+            raise exceptions.BadRequest(detail='Empty JSON')
         try:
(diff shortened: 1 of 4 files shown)


tomchristie:
Good stuff, yup. I'd rather that we don't have exception subclasses for this, but instead just create a general `BadRequest` for 400 errors. (Similar to the `ValidationError` class) We'll then pass explicit arguments to it.

pavelholvi:
Shall we perhaps use `BadRequest` where we previously used `ValidationError` and have only 1 exception with status 400?

tomchristie:
Good stuff, thanks!

PR #299, opened 2017-09-25
change the README docs

kinabalu:


audiolion:
I am not sure why you would change the name to `IsAuthorized`?

kinabalu:
because permissions are generally a function of authorization, and not authentication.

audiolion:
I'd argue that in the authentication module you create permissions that are self describing, e.g. `IsAdmin`, `IsOwner`, `IsAuthenticated`, etc. `IsAuthorized` would make sense if you were inheriting that permission and overriding the `has_permission` class, but you aren't you just write a class that implements `has_permission` and returns a boolean and pass the permission in.

tomchristie:
Nope, as @audiolion points out, the permission class is "You can do this *if* you are authenticated".

> if I saw IsAuthorized permission attached to a view function, I would have no idea what it means to be authorized, if it was IsAuthenticated then I know immediately you need to be logged in to access the view.

Exactly.

PR #300, opened 2017-09-25
Fix doc error, change Renderer to Parser.

grangerp:


tomchristie:
😱  How'd I miss that?

Thanks! 😄

Issue #244, opened 2017-08-18
Typesystem refinements


tomchristie:
Pulling these out of `TODO` comments, and into an issue...

* Error on unknown attributes
* allow_blank?
* input_type (check type at start and allow, coerce, .native)
* default=empty
* check 'required' exists in 'properties'
* smarter ordering
* extra_properties=False by default
* Overriding errors
* Blank booleans as False?

Bit of a scattergun list, but worth tracking for now.

tomchristie:
Obsoleted by #400

PR #302, opened 2017-09-27
Treat an empty string boolean type as False

mammuth:
This makes blank boolean strings handled as a False boolean instead of raising an TypeSystemError saying it's an invalid boolean.

As mentioned in #244.


tomchristie:
👍

PR #303, opened 2017-09-29
Update Building Websites notes in README.md

martinpeck:
After suggesting that the README.md needed to be updated because of the changed behaviour w.r.t. content types and HTML responses, I thought it might be helpful to make the changes myself and send a PR.

Here, I've described three options for returning HTML (annotation, explicit Response objects, and AFTER/BEFORE request hooks).

### DIFF
--- a/README.md
+++ b/README.md
@@ -538,11 +538,12 @@ responses, using [Jinja2](http://jinja.pocoo.org/).
 **app.py:**
 
 ```python
-from apistar import Route
+from apistar import Route, annotate
 from apistar.interfaces import Templates
 from apistar.frameworks.wsgi import WSGIApp as App
+from apistar.renderers import HTMLRenderer
 
-
+@annotate(renderers=[HTMLRenderer()])
 def hello(username: str, templates: Templates):
     index = templates.get_template('index.html')
     return index.render(username=username)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -561,9 +562,15 @@ settings = {
...
-content type. You can override this by returning a `Response`, including an
-explicit `Content-Type` header.
+Returning a string response from a view will default to using the `text/json`
+content type. This means you will need to override this so that your HTML views
+use the `text/html` content type in their responses. There are several ways you can do this,
+including:
+
+1. by annotating your handler function so that it uses the `HTMLRenderer` (as shown above)
+2. by returning a `Response`, including an explicit `Content-Type` header (see [Renderers](#renderers) section for an example)
+3. by using the `BEFORE_REQUEST` and `AFTER_REQUEST` hooks, which fire before and after all requests
```
tomchristie: Let's drop this last point. Yes, you *could* use them for that purpose, but it's a convoluted way around.

martinpeck: Agreed. I wasn't sure if I should include it or not. I assume it would only make sense to use BEFORE/AFTER if you're writing a 100% HTML site...at which point, is apistar the framework you're likely to be using?

tomchristie: In that case you'd want to set the default renderer to HTML in the global settings (rather than in annotations)

martinpeck:
I wondered if a couple of subclasses of `Response`, such as `HtmlResponse` and `JsonResponse`, would keep the logic within the handler while also reducing he amount of typing. 

So, the handler would look something like...

```python
def welcome():
    content = "<html>...</html>"
    return HtmlResponse(content)
```

(`HtmlResponse` would accept all the same named arguments as `Response`)

`JsonResponse` would do the same as `Response`, except in situations where the default renderer had been changed (i.e. `HtmlResponse` and `JsonResponse` would override any other response type settings).

tomchristie:
@martinpeck Had already considered that too, and yes, 100% we want that! 😄

tomchristie:
(Aside let's uppercase, as `HTMLResponse`, and `JSONResponse`)

PR #304, opened 2017-09-29
Add pool_recycle and pool_size to mysql sqlalchemy

nydr:
Add pool_recycle and pool_size to mysql sqlalchemy backend.

MySQL has a setting `wait_timeout` which will drop connections after being idle for a certain time. This setting is defaults to 8 hours but some distributions default it to 10 minutes. If a connection is dropped by MySQL without SQLAlchemy expecting it it will cause the client to raise an exception with the message (2006, 'MySQL server has gone away'). Setting `pool_recycle` to less than `wait_timeout` is a common way in SQLAlchemy to work around this as it will cause SQLAlchemy to replace any connections older than this value.

### DIFF
--- a/apistar/backends/sqlalchemy_backend.py
+++ b/apistar/backends/sqlalchemy_backend.py
@@ -22,6 +22,9 @@ class SQLAlchemyBackend(object):
         kwargs = {}
         if url.startswith('postgresql'):  # pragma: nocover
             kwargs['pool_size'] = database_config.get('POOL_SIZE', 5)
+        elif url.startswith('mysql'):  # pragma: nocover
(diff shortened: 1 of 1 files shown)


nydr:
I set the default to 1 hour as it seems like a reasonably sane value, but I'll be just as happy keeping it as unset by default and letting users configure it via settings.

mehrdadmhd:
Saw this pull request after [mine](https://github.com/encode/apistar/pull/403) ...
I suffered from closed connections too using MariaDB.

tomchristie:
Obsoleted by #400 - we should start pushing for ORM integration etc. to be third party packages. (Tho typically you only need that if you want eg. implicit commit/rollback dealt with by event hooks)

PR #306, opened 2017-10-02
Add "yes/no" as valid booleans

imanhodjaev:
Sometimes people also use yes/no as boolean value so I added them to the list of possible boolean values.


imanhodjaev:
Closing this PR because there is no more `typesystem` module.

PR #312, opened 2017-10-04
typesystem.Object: Add errors for unknown property keys

mammuth:
With this PR we now raise field errors for unknown fields of an object. 

Since we use stuff like __state_ or __sa_instance_state_ internally, field names that start with an underscore won't raise errors - does this make sense? 😊 

As stated in #244


tomchristie:
This is resolved in #400, since we now get `additional_properties=[True|False]`

PR #323, opened 2017-10-12
Python 3 style implicit inherit from object

yunti:
Bit of a style choice, but this makes it clearer this is a python 3 codebase.  (Now also the style used in Django)


tomchristie:
👍  Yup.

Issue #236, opened 2017-08-08
CLI Sub-Commands 


jwbixby:
Obviously no ones top priority currently, but I was playing around a bit and got a pretty simple MVP of sub-commands(nested commands) working. Please let me know if this is a feature that you'd like to see in apistar. If it is, I'll clean this up and write some tests.

https://github.com/jwbixby/apistar/pull/1/files

My intentions are to simplify the commands interface by allowing namespacing. Users can create groups of commands that can be nested within each other. Very similar to the way `Include` works with routes.

Previously, all the backend commands were prefixed with their respective library names. `apistar django_makemigrations`, `apistar django_migrate`, etc... Using subcommands, we are able to join these together into `apistar django --help`, `apistar django makemigrations`, `apistar django migrate`. 

This is building off of the 0.2.0 version branch.
```python
django_commands = SubCommand(
    'django',
    [
        Command('makemigrations', make_migrations),
        Command('migrate', migrate),
    ],
)

app = App(commands=[django_commands, Command('other', other), ... ])
```

The current output in the MVP is not this clean, but potentially could look like:

```
bash-3.2$ apistar --help
Usage: apistar [OPTIONS] COMMAND [ARGS]...

  API Star

Options:
  --version  Display the `apistar` version number.
  --help     Show this message and exit.

Commands:
  django        Perform Django framework commands (makemigrations, migrate, etc)
  new            Create a new project in TARGET_DIR.
  run              Run the current app.
  schema      Output an API Schema.
  test           Run the test suite.
```
```
bash-3.2$ apistar django --help
Usage: apistar django [OPTIONS] COMMAND [ARGS]...

  Django Framework Commands. Perform various commands from the Django toolkit...

Options:
  --help     Show this message and exit.

Commands:
  makemigrations        Create new migrations files
  migrate            Execute the migrations
  run              Run the current app.
```

I also played around with adding `help` and `description` to the parsers by adding kwargs to the Command and SubCommand classes.

```python
commands = [
    Command('makemigrations', make_migrations, help='Perform migrations', description='Perform migrations')
]
```

jwbixby:
Also, not sure where I can best report this. I am unable to verify my email on `discuss.apistar.org`. I receive the message below after clicking "Click here to activate your account".

```
[
"BAD CSRF"
]
```

tomchristie:
>  I receive the message below after clicking "Click here to activate your account".

[...]

PR #328, opened 2017-10-16
RFC: Sub-Commands

jwbixby:
Issue: https://github.com/encode/apistar/issues/236

## Introduction

This RFC goes over a potential implementation of Sub-Commands. The goal of the Sub-Commands feature is to simply the command line interface by allowing commands to be grouped and documented.

The following section merely introduces the implementation. Readers should reference the updated `README.md` documentation for examples on how to apply these features.

## Implementation

This RFC introduces two main pieces to achieve the features goals.

1. `SubCommand` - The ability to group commands together under a parent command.
2. Documentation - The ability to add inline documentation that enhances the the command line interface with more information and details.

### SubCommand

The `SubCommand` class is similar to the `Environment` class in the sense that it allows users to define class properties that will be used later when the user instantiates it. This allows the user to define properties that should work in most cases, but later these values can be overridden.

One situation where a user might want to override these values is when this SubCommand class is defined by a third party(e.g. apistar-sqlalchemy). Users can easily override the predefined `name` property to any name they'd like.

### Documenation

I believe this section is best explained by the documentation added to README.md. TLDR; By standardizing how documentation is written in the doc string, we can provide a better interface in the command line.

---

After preparing this, I did some thinking on how this could be applied to Routes. I left some notes on a branch in my fork that shows how these could be applied in a very similar way and help document nested routes. I'll leave a link here, but its out of scope for this PR. https://github.com/jwbixby/apistar/blob/jb_future_of_routes/route_explaination.md


README.md
```diff
@@ -294,7 +297,8 @@ app = App(routes=[
 
 ### Reversing URLS
 
-Rather than build URLs by hand, it is possible to generate URLs based on the endpoint, using `reverse_url()`.
+Rather than build URLs by hand, it is possible to generate URLs based on the
+endpoint, using `reverse_url()`.
```
audiolion: In markdown, you shouldn't insert your own linebreaks, leave paragraphs all on the same line

README.md
```diff
@@ -320,10 +324,19 @@ app = App(routes=[
 ])
 ```
 
-## Command Routing
+---
+
+# Commands
+
+In addition to routing URLs, you can also route commands, to make
+additional functionality available directly to the command line client.
```
audiolion: remove linebreak

README.md
```diff
@@ -345,6 +358,126 @@ commands = [
...
+* `name` - A string that will identify the group of commands. If not
+specified, the `SubCommand` will automatically generate a name based on the class
+name by truncating `"Commands"`. For example, `UserCommands` will be under the
+command named `user`.
+* `commands` - A list of `Command`s or `SubCommand`s. These will be the nested commands.
+
+### Example
+
+The following `app` configuration will have 4 top level commands: `user`,
+`override_user`, `import`, and `other_command`.
```
audiolion: Remove linebreak

README.md
```diff
@@ -345,6 +358,126 @@ commands = [
...
+```
+
+### Documenting Commands
+
+Commands are documented by the doc string on the extended `SubCommand` class and
+command functions. Both follow the same format of a short description on the first
+line, an empty line, and a longer description starting on the third line. The
+longer description can be any length. Failure to separate the short and long
+description with an empty line will result in only the long description being
+specified.
```
audiolion: remove above linebreaks

jwbixby:
I love the new `Document` idea for `0.4.0`. Keep up the great work! Can't wait for this framework to be stable.

PR #330, opened 2017-10-17
JSON parser encoding error raises a 500

ysavary:
If the json content is not properly encoded (apistar assumes that json are in utf-8), raise also a 400 'Invalid JSON'


tomchristie:

PR #331, opened 2017-10-17
Taking in account @annotate(parsers=...)

ysavary:
It seems that it's missing the code that take in account '@annotate(parsers=...'


tomchristie:
Obsoleted by #400.

PR #334, opened 2017-10-18
Added a test that demonstrates the 500 error with a wrong 'Accept' header

ysavary:
Just added a test that demonstrates the 500 error when an wrong 'Accept' header is passed

### DIFF
--- a/apistar/hooks.py
+++ b/apistar/hooks.py
@@ -65,6 +65,7 @@ def render_response(handler: Handler,
         renderer = negotiate_renderer(accept, renderers)
         if renderer is None:
             raise exceptions.NotAcceptable()
+            # return http.Response(status=406)
         content = injector.run(renderer.render, {'response_data': data})
         if isinstance(content, http.Response):
             return content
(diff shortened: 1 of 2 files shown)


apistar/hooks.py
```diff
@@ -65,6 +65,7 @@ def render_response(handler: Handler,
         renderer = negotiate_renderer(accept, renderers)
         if renderer is None:
             raise exceptions.NotAcceptable()
+            # return http.Response(status=406)
```
drpoggi: Looks like this slipped in by mistake :)

ysavary: This commented line will make the test pass but it's certainly not the right way to fix the issue.

drpoggi:
I think you may want to take a second look at this, `test_text_with_wrong_accept_header` is failing in the Travis build.

ysavary:
@drpoggi i'm sorry, i tried but i don't know how to fix properly this issue. I added this failing test "test_text_with_wrong_accept_header" to demonstrate the issue.

drpoggi:
Ah I see. Is there an existing Github issue tracking the error you're getting?

ysavary:
It seems that NotAcceptable exception is not handled properly in this rendered context. A 2nd uncaught exception is propagated up to the root.

Returning a 406 response from here works but, in this case, i don't know how to call the error_handler to render the correct error response.

tomchristie:

PR #335, opened 2017-10-19
デフォルト値

243083df:


audiolion:
This might be [intentionally done] (http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) due to python's mutable default arguments

243083df:
@audiolion Strings immutable after all;

audiolion:
| @audiolion Strings immutable after all;

👍  learned something new every day! 

私は理解する、ありがとう

randvis:
My wild guess is it's author's common practice to distinguish **string argument not assigned by caller** from **string argument is assigned as empty by caller**. So the default value of string argument is None.
However in this snippet, it's not necessary to distinguish between these so the code looks a bit weird.

Other two cases in which default value of a string is None are:
https://github.com/encode/apistar/blob/master/apistar/core.py#L273 
```
class Include(collections.abc.Iterable):
    def __init__(self,
                 path: str,
                 routes: typing.Sequence[typing.Union[Route, 'Include']],
                 namespace: str=None) -> None:
        self.path = path
        self.routes = routes
        self.namespace = namespace
```

https://github.com/encode/apistar/blob/master/apistar/exceptions.py#L4
```
    def __init__(self,
                 detail: Union[str, dict]=None,
                 cls: type=None,
                 code: str=None) -> None:

        if cls is not None and code is not None:
            errors = getattr(cls, 'errors')
            detail = errors[code].format(**cls.__dict__)

        self.detail = detail
        super().__init__(detail)
```

243083df:
@randvis In https://github.com/encode/apistar/blob/master/apistar/core.py#L27
None is a valid value and compared against it in
https://github.com/encode/apistar/blob/master/apistar/core.py#L82

tomchristie:
Fair enough, yup!

PR #336, opened 2017-10-19
Fix comment error for render_response

randvis:
`render_response()` is a after request hook, not before.


tomchristie:
Obsoleted by #400, which is very close to being merged in now.

Thanks for this, and sorry for having been away from the project for so long.

PR #337, opened 2017-10-20
性能を上げる

243083df:


Rdbaker:
Could you provide benchmarks for how much this increases performance? It's always good to have a baseline then show improvement

243083df:
In this benchmark, new subtype parsing 3 times faster then current.
Benchmark code:
```python
import re
import time

a = ["text/plain", "applicationjson", "text/html"]
subtype = re.compile('/[^/]*$')
media_type = "application/*"

start = time.time()
for j in range(1000000):
    for i in a:
        match = re.sub(subtype, '/*', i)
        if media_type == match:
            pass
end = time.time()
print("Old")
print(end - start)


start = time.time()
for x in range(1000000):
    typ = media_type[:-2]
    for i in a:
        try:
            match = i[:i.rindex('/')]
        except ValueError:
            match = i
        if typ == match:
            pass
end = time.time()
print("New")
print(end - start)
```

Result:
```
Old
6.139465808868408
New
1.8832921981811523
```

PR #342, opened 2017-10-27
Create MongoDB pymongo backend

humrochagf:


mammuth:
You can run `scripts/lint` and `scripts/test`  to see what causes the Travis build to fail :wink:

tomchristie:
Great stuff, tho we'd certainly want it as a third party package, rather than built-in.
(I think we'll eventually do the same for the existing backends too)

humrochagf:
cool, i was thinking about it when i see all the options of ORM`s for SQL and NOSQL databases
I will extract that to an external package
Thanks for the tip

PR #345, opened 2017-10-27
Fix new lint issue in flake8 3.5.0

randvis:
flake8 new version 3.5.0 doesn't allow bare `Except` any more.

PR #352, opened 2017-11-03
Process default values with type

GomZik:
# Example of usage
```python
import typing
from apistar import typesystem, environment


class Database(typing.NamedTuple):
    HOST: str
    PORT: int
    USER: str
    PASSWORD: str
    NAME: str

class DSN(typesystem.String):
    def __new__(cls, *args, **kwargs):
        value = super().__new__(cls, *args, **kwargs)
        # Parse DSN value
        return Database(host, port, user, password, name)
```

Next I define Environment

```python
class Env(environment.Environment):
    properties = {
        'DATABASE': typesystem.newtype(DSN, default='postgresql://scott:tiger@localhost/test')
    }

env = Env()
print(env['DATABASE'])
```
Current result if no environment variable passed is:
`postgresql://scott:tiger@localhost/test`

Expected by me:
`Database(HOST='localhost', PORT=5432, USER='scott', PASSWORD='tiger', NAME='test')`

Of course, I can use next definition
```python
typesystem.newtype(DSN, default=DSN('postgresql://scott:tiger@localhost/test'))
```

OR

```python
typesystem.newtype(DSN, default=Database('localhost', 5432, 'scott', 'tiger', 'test'))
```

But I think, that my variant is better. What do you think?

### DIFF
--- a/apistar/typesystem.py
+++ b/apistar/typesystem.py
@@ -176,7 +176,7 @@ class Object(dict):
             except KeyError:
                 if hasattr(child_schema, 'default'):
                     # If a key is missing but has a default, then use that.
-                    self[key] = child_schema.default
+                    self[key] = child_schema(child_schema.default)
                 elif key in self.required:
                     exc = TypeSystemError(cls=self.__class__, code='required')
                     errors[key] = exc.detail


mammuth:
You can run `scripts/lint` and `scripts/test`  to see what causes the Travis build to fail :wink:

GomZik:
It will work when #345 will be merged

tomchristie:
Resolved in #400 (See https://discuss.apistar.org/t/a-peek-at-the-upcoming-version/412 for a bit of info on how the new types look, and an example including `default=...`)

PR #356, opened 2017-11-12
Handle 406 not acceptable response

randvis:
Fix #334. Return correct response when Accept header cannot be satisfied.


randvis:
All test cases are passed. Travis checking should work after #345 is merged.

tomchristie:

Issue #332, opened 2017-10-18
AttributeError: 'Headers' object has no attribute 'get'


naoko:
`Python 3.5.2`
`apistar==0.3.9`

I'm getting `AttributeError: 'Headers' object has no attribute 'get'` from `apistar/components/wsgi.py` in get_header.

This works on a few environments but it error out on others. I compare `pip freeze` output and all environment are identical.
On working instances, python versions are 3.5.3 or 3.5.4. 

and when I change the line from `.get` to `.__getitem__` all works.
```
def get_header(name: ParamName, headers: http.Headers):
    return headers.__getitem__(name.replace('_', '-'))
```
Any advice would be appreciated. 

UPDATE:
I've try to run on other system that has python 3.5.2 and got the same error. Now I'm convinced that this is issue on 3.5.2. Can we test with `hasattr`

ratulotron:
Faced the same issue. Fixed by upgrading to Python 3.6.

anseljh:
I ran into this one too, and it was rather sneaky. I'd been developing on 3.6.4, and then deployed to a server with 3.5.2. Instant breakage, even though `pip freeze` output was identical.

Relevant portion of my traceback:
```
  File "/home/username/.local/share/virtualenvs/api-gjVyAMqI/lib/python3.5/site-packages/apistar/frameworks/wsgi.py", line 134, in __call__
    response = self.http_injector.run_all(funcs, state=state)
  File "/home/username/.local/share/virtualenvs/api-gjVyAMqI/lib/python3.5/site-packages/apistar/components/dependency.py", line 141, in run_all
    ret = step.func(**kwargs)
  File "/home/username/.local/share/virtualenvs/api-gjVyAMqI/lib/python3.5/site-packages/apistar/frameworks/wsgi.py", line 130, in __call__
    response = self.http_injector.run_all(funcs, state=state)
  File "/home/username/.local/share/virtualenvs/api-gjVyAMqI/lib/python3.5/site-packages/apistar/components/dependency.py", line 141, in run_all
    ret = step.func(**kwargs)
  File "/home/username/.local/share/virtualenvs/api-gjVyAMqI/lib/python3.5/site-packages/apistar/components/wsgi.py", line 67, in get_header
    return headers.get(name.replace('_', '-'))
AttributeError: 'Headers' object has no attribute 'get'
```

Would love to see #364 get merged. 👏

tomchristie:
Thanks for raising this... resolved in #400.

PR #364, opened 2017-11-22
Fixing issue 332: No .get() method when Python 3.5.2 is used

michwill:
When Python 3.5.2 (or lower?) is used, `apistar.http.Headers` and `apistar.http.QueryParams` make many of tests fail because there is no `.get()` method (https://github.com/encode/apistar/issues/332). In newer versions of Python `.get()` method comes with `typing.Mapping` which these classes subclass.

This PR creates a mixin class which adds a `get` method if it doesn't exist yet. However, if it does (e.g. in Python 3.6), the standard one (from `typing.Mapping`) is used.

### DIFF
--- a/apistar/http.py
+++ b/apistar/http.py
@@ -2,6 +2,7 @@ import collections
 import io
 import typing
 from urllib.parse import urlparse
+from apistar.types import MappingGetMixin
 
 Method = typing.NewType('Method', str)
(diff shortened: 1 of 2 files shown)


tomchristie:
Good call - including this in #400

tomchristie:
Thanks!

PR #367, opened 2017-11-29
Add required argument APP to the run subcommand

### LINKED ISSUE #350: Running app from other than the working directory
Something like

```bash
apistar run ./api
```

to run the APIStar instance inside that directory, so users don’t need to change directory first. It might be a good idea to allow an entry point other than `app.py` as well.

pslacerda:
This PR is an effort to fix  #350, is somewhat related to the PR #259.

To be able to run the application from any directory is valuable and possible in many other frameworks. For instance gunicorn:

    gunicorn webapp.somemodule:app \
        --worker-class=meinheld.gmeinheld.MeinheldWorker. 

I addressed this issue in a very similar manner, now `apistar run` requires a positional APP argument which make possible to run `App` from anywhere:

    apistar run myapp.gateway.app:app

I'm concerned with the `import_app()` because the module unloading was mainly untested.

I really liked of the command line approach using MyPy and dependency injection, even that it turned very difficult for me to understand.

Congratulations for this nice framework. We are attempting to run it on production.

Maybe this PR can fix this annoying issue.

### DIFF
--- a/apistar/__init__.py
+++ b/apistar/__init__.py
@@ -10,7 +10,6 @@ from apistar import exceptions
 from apistar.core import Command, Component, Include, Route, annotate
 from apistar.frameworks.cli import CliApp
 from apistar.http import Response
-from apistar.interfaces import App
 from apistar.test import TestClient
 from apistar.types import Settings
 from apistar.utils import import_app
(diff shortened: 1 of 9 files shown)


pslacerda:
Probably is better something like the following

```python
def command(var: (conv_func, 'default')):
    pass
```

Where `conv_func()` convert `str` into the wanted object. The optional `'default'` is passed in the absense of  user supplied argument. For example

```python
def run(app: (import_app, 'app:app')):
    pass
```
Currently it is missing support to `type` or conversor functions like argparse `type` parameter.

tomchristie:
Closing this off, given 0.4.

PR #371, opened 2017-12-04
Update Readme.md

trthhrtz:
Add corrections to the readme - dev. section


tomchristie:
Good call. I'll close this off because I don't want to create merge errors with #400, but yes we should suggest cloning from https:

PR #378, opened 2017-12-26
factorize coerce or fail code

mgaitan:
I factorized a repeated pattern because of the following reasons: 

- [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)
- I need to manually validate a "create" endpoint that must receive different payploads depending of a parameter on the url. So I could do something like this

```python
Route('/create/{slug}', 'POST', create)
... 
def create(slug: Product, data):
     product = coerce_or_fail(PRODUCT_TYPES[slug], data)
     ... 
```

### DIFF
--- a/apistar/components/dependency.py
+++ b/apistar/components/dependency.py
@@ -549,6 +549,34 @@ class CliResolver(Resolver):
         return kwargs[name]
 
 
+def coerce_or_fail(coerce,
+                   value,
+                   exception: Exception=exceptions.ValidationError,
+                   name=None):
+    """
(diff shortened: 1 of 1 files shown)


mgaitan:
travis' failure seems to have no relation with my PR. 

```
apistar/backends/django_orm.py:53:5: E722 do not use bare except'
apistar/backends/sqlalchemy_backend.py:44:5: E722 do not use bare except'
The command "scripts/lint" exited with 1.
```

tomchristie:
No longer exists in #400.

Issue #377, opened 2017-12-24
Map Multiple HTTP Methods to same view function


audiolion:
```python
account_routes = [
    Route('/{uuid}/', 'PUT', views.update_account),
    Route('/{uuid}/', 'PATCH', views.update_account),
]
```

possible to have two methods like PUT/PATCH map to the same view function?

mgaitan:
`Route` should accept a string or a tuple of string as methods 

```
account_routes = [
    Route('/{uuid}/', ('PUT', 'PATCH'), views.update_account),
]
```

audiolion:
This should work but the way its coded it doesn't:

https://github.com/encode/apistar/blob/973c6485d8297c1bcef35a42221ac5107dce25d5/apistar/components/router.py#L40

it wraps what is passed in a list, so it would pass `[('PUT', 'PATCH')]` and then werkzeug fails to parse a tuple object when it calls:

```python
self.methods = set([x.upper() for x in methods])
```

tomchristie:
Closing as per #386

PR #386, opened 2018-01-22
Map Multiple HTTP Methods to same view function

bahattincinic:
I tried to fix this error in a `apistar/apistar/components/router.py` file. I got an error.
Because. `coreapi `doesn't support multiple methods to the same `coreapi.Link`

issue: #377.

### DIFF
--- a/apistar/core.py
+++ b/apistar/core.py
@@ -71,7 +71,7 @@ def flatten_routes(routes: typing.Sequence[typing.Union[Route, Include]],
             name = namespace_prefix + name
             if isinstance(method, (list, tuple)):
                 for _method in method:
-                    path_name = f"{_method}_{name}"
+                    path_name = "%s_%s" % (_method, name)
                     route = Route(path, _method, view, path_name)
                     flattened_routes.append(route)
             else:


apistar/core.py
```diff
@@ -69,8 +69,14 @@ def flatten_routes(routes: typing.Sequence[typing.Union[Route, Include]],
             path, method, view, name = item
             path = path_prefix + path
             name = namespace_prefix + name
-            route = Route(path, method, view, name)
-            flattened_routes.append(route)
+            if isinstance(method, (list, tuple)):
+                for _method in method:
+                    path_name = f"{_method}_{name}"
```
povilasb: F-strings only work on python 3.6. See: https://www.python.org/dev/peps/pep-0498/
But it seems like `apistar` aims to support python 3.5 too.

bahattincinic: nice catch. thanks.

audiolion:
@tomchristie confused on this, so does that mean if we want a `PUT` and `PATCH` for an `edit_user` route, we need to do:

```python
def edit_user(username):
    pass

routes = [
    Route('/{username}', 'PUT', edit_user),
    Route('/{username}', 'PATCH', edit_user),
]
```

seems like apistar should be able to take an iterable route and replicate it internally to the multiple routes above to provide a short-hand syntax for developers

tomchristie:
@audiolion The issue is that we map our routes onto a representation that can be used for schema generation. For the purposes of schema output or schema-driven clients interacting with the API each of the `PUT` and `PATCH` operations need to be treated distinctly, with *a unique name on each*, so we can differentiate which action the client should perform.

Ie. the schema needs to be specific as to what method something like `client.action('edit_user', ...)` should use in the request.

PR #398, opened 2018-03-06
Handle unexpected arguments in server['message']

ytjohn:
On some servers (ubuntu 16.04), if you run apistar with localhost, then message['server'] will return 4 values instead of 2.

Example: 
```
$ host localhost
localhost has address 127.0.0.1
localhost has IPv6 address ::1
apistar run --host localhost --port 5010
```

This can be worked around by specifying 127.0.0.1 instead of localhost, or editing /etc/hosts, but a more reliable solution is to limit the returned range within the code.

Traceback/debuging
```
ValueError: too many values to unpack (expected 2)
File "/usr/share/python/certainmesh/lib/python3.6/site-packages/apistar/components/umi.py", line 19, in get_url
Open an interactive python shell in this framehost, port = message['server']

>>> print(message.get('server'))
('::1', 5010, 0, 0)
>>> host, port = message['server'][0:2]
>>> host
'::1'
>>> port
5010
```

### DIFF
--- a/apistar/components/umi.py
+++ b/apistar/components/umi.py
@@ -16,7 +16,7 @@ def get_method(message: UMIMessage):
 
 def get_url(message: UMIMessage):
     scheme = message['scheme']
-    host, port = message['server']
+    host, port = message['server'][0:2]
     path = message['path']
 
     if (scheme == 'http' and port != 80) or (scheme == 'https' and port != 443):


tomchristie:
I’d first want to understand:

What are the other two parts of the value? Why is it populated as 4 values?

ytjohn:
I don't know exactly where or how far down apistar is getting the address for localhost. However, this is actually an ipv6 issue. I just happened
to come across it because the servers I'm deploying to expose it with localhst. A lot of linux servers add 127.0.0.1 and ::1 for localhost in /etc/hosts. Somewhere, one of apistar's dependencies is using a "socket.get*" call. Essentially, if the address your server is running on has an ipv6 address, you get two extra argument called "flowinfo" and "scopeid". 

https://docs.python.org/3/library/socket.html#socket-families

> A pair (host, port) is used for the AF_INET address family, where host is a string representing either a hostname in Internet domain notation like 'daring.cwi.nl' or an IPv4 address like '100.50.200.5', and port is an integer.

> For AF_INET6 address family, a four-tuple (host, port, flowinfo, scopeid) is used, where flowinfo and scopeid represent the sin6_flowinfo and sin6_scope_id members in struct sockaddr_in6 in C. For socket module methods, flowinfo and scopeid can be omitted just for backward compatibility. Note, however, omission of scopeid can cause problems in manipulating scoped IPv6 addresses.

If the address you're binding to only has an ipv4 address, it looks like this:

    >>> socket.getaddrinfo("localhost", 8080, 0, 0, socket.IPPROTO_TCP, 0)
    [(2, 1, 6, '', ('127.0.0.1', 8080))]

If the address you're binding to has both ipv4 and ipv6, you get this:

    >>> socket.getaddrinfo("localhost", 8080, 0, 0, socket.IPPROTO_TCP, 0)
    [(10, 1, 6, '', ('::1', 8080, 0, 0)), (2, 1, 6, '', ('127.0.0.1', 8080))]

What I observed is that I would randomly get one or the other simply by refreshing an endpoint or running curl in a loop. Once apistar
went to ipv6, I would get the traceback below. I used the Werkzeug traceback console in the browser to find the issue.

```
Traceback (most recent call last):
  File "/opt/certainmesh/lib/python3.6/site-packages/apistar/frameworks/asyncio.py", line 125, in __call__
    response = await self.http_injector.run_all_async(funcs, state=state)
  File "/opt/certainmesh/lib/python3.6/site-packages/apistar/components/dependency.py", line 430, in run_all_async
    ret = step.func(**kwargs)
  File "/opt/certainmesh/lib/python3.6/site-packages/apistar/frameworks/asyncio.py", line 121, in __call__
    response = await self.http_injector.run_all_async(funcs, state=state)
  File "/opt/certainmesh/lib/python3.6/site-packages/apistar/components/dependency.py", line 430, in run_all_async
    ret = step.func(**kwargs)
  File "/opt/certainmesh/lib/python3.6/site-packages/apistar/components/umi.py", line 19, in get_url
    host, port = message['server']
ValueError: too many values to unpack (expected 2)
```
 
And here is the state var from debug:

```
File "/opt/certainmesh/lib/python3.6/site-packages/apistar/frameworks/asyncio.py", line 125, in __call__
response = await self.http_injector.run_all_async(funcs, state=state)
[console ready]
>>> print(state)
{'message': {'channel': 'http.request', 'server': ('::1', 5010, 0, 0), 'client': ('::1', 57020, 0, 0), 'scheme': 'http', 'http_version': '0.0', 'method': 'GET', 'path': '/', 'query_string': b'', 'headers': [[b'x-real-ip', b'192.168.145.137'], [b'x-forwarded-for', b'192.168.145.137'], [b'host', b'localhost:5010'], [b'connection', b'close'], [b'cache-control', b'max-age=0'], [b'upgrade-insecure-requests', b'1'], [b'user-agent', b'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/65.0.3325.162 Safari/537.36'], [b'accept', b'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8'], [b'accept-encoding', b'gzip, deflate, br'], [b'accept-language', b'en-US,en;q=0.9'], [b'cookie', b'session_id=cab6df82-2d78-11e8-873b-005056aeba8d']]}, 'channels': {'reply': <uvicorn.protocols.http.ReplyChannel object at 0x7f77c42499d8>}, 'handler': <function homepage at 0x7f77c745aea0>, 'kwargs': {}, 'exc': ValueError('too many values to unpack (expected 2)',), 'response_headers': Headers([])}
>>> 
```

tomchristie:
Going to close this off given that we're now ASGI based. Happy to see it re-opened if it *is* still an issue tho.

PR #400, opened 2018-03-13
Towards 0.4

tomchristie:
Lots of big changes here. I'll keep the synopsis to a minimum...

* Rewrite type system. (Don't use type subclasses, just plain old python objects please.)
* Rewrite components. (Much simpler. Components as plain objects allows per-component configuration, rather than everything in a big global settings dict.)
* Add event hooks.
* ASGI becomes the asyncio interface, since it's latest re-write now supports our requirements.
* OpenAPI to become the default schema representation.
* Drop `coreapi` and `coreschema` - Schema generation and client library is now part of this package.
* Drop built-in support for ORMs, push towards third-party instead.

Here's the [README](https://github.com/encode/apistar/tree/level-up)

TODO:

- [x] Format support, eg. 'datetime' validation.
- [x] Enforce location=`query|path|body`
- [x] Enforce path fields as required, and matching uritemplate.
- [x] Remove uritemplate dependency.
- [x] Support `Type`. 
- [x] Deal with non-unique names in Routing (asserts at document level)
- [x] Include `Type` and annotation information on schema.
- [x] Allow `Type` subclasses to be used in validations.
- [x] ASyncIO Support, with ASGI
- [x] Static Files support for ASGI 
- [x] Update `uvicorn` for ASGI Support - See https://github.com/encode/uvicorn/pull/63
- [x] Resolve issue with dependency injection not persisting parameters between functions.
- [x] Clean up TODOs in docs.

**Later**:

- [ ] Add `url` to Section.
- [ ] Guard against uvicorn calling into old-style interface, and yell loudly.
- [ ] Response schemas in OpenAPI.
- [ ] Consider `types`/`fields` instead of `types`/`validators`.
- [ ] Bool and null coercions for query string and form data.
- [ ] Use ConfigurationError, not `assert`
- [ ] Subclassing `Type` and handling `allOf` relationships in JSON schema.
- [ ] Reintroduce API docs.
- [ ] Reintroduce command routing.
- [ ] Test coverage & fixups.
- [ ] Third-party Flask, Django variants.
- [ ] Check for configuration errors when instantiating router from Document.
- [ ] Use `url` not `base_url` for decode option from HTTPTransport.
- [ ] Support `errors` override on type system.
- [ ] Client should validate path/query/body parameters are appropriate primitive types only.


apistar/compat.py
```diff
@@ -0,0 +1,32 @@
...
+import sys
+
+if sys.version_info < (3, 6):
+    dict_type = collections.OrderedDict
+else:
+    dict_type = dict
+
+
+try:
+    # Ideally we subclass `_TemproraryFileWrapper` to present a clear __repr__
```
audiolion: s/_TemproraryFileWrapper/_TemporaryFileWrapper

apistar/server/injector.py
```diff
@@ -0,0 +1,108 @@
...
+
+from apistar.exceptions import ConfigurationError
+
+
+class Component():
+    resolves = None
+
+    def identity(self, parameter: Parameter):
+        """
+        Each component needs a unique identifier string that we use for lookups
```
audiolion: perhaps mention that it is a unique, case-insensitive name?

tomchristie: Thanks for the review! The name isn't actually case-insensitive, I'm just using the lowercase for stylistic reasons.

apistar/compat.py
```diff
@@ -0,0 +1,32 @@
+import collections
+import sys
+
+if sys.version_info < (3, 6):
+    dict_type = collections.OrderedDict
+else:
+    dict_type = dict
```
GomZik: Not sure, that we can assume, that dict has ordered keys since 3.6

tomchristie: dict is ordered in cpython 3.6 as an implementation detail, and guaranteed ordered in 3.7+

We don't *strictly* need ordering, it's only for stylistic purposes of eg. returning JSON outputs that are always in the same way as the fields in the codebase, so it's not a massive problem if there are a few edge cases of platforms that *could* get unordered outputs, but I'd be happy if there's any stricter checking that we can do, to also check if we're on `cpython` or some other implementation.

RunningMartin:
Hai,It's nice to hear that the author started to update this framework again. I have an immature suggestion,I hope the author can refer to they.
- Ability to create an app and then add routes or other configurations by some initialization method
```
app = App()
app.init_app(routes=routes,.......)
```

tomchristie:
@jgirardet Yup will probably go with that for eg. `path_fields` on the links.

tomchristie:
@RunningMartin What's the motivation there?

RunningMartin:
@tomchristie I only want to delay the initialization of the app.Maybe I was affected by the flask.

jgirardet:
I don't see in TODO or in the code (App ?) the way to add third party component like we did before with (components = components during App instanciating process) 
Does it change ?

apistar/formats.py
```diff
@@ -0,0 +1,97 @@
+import datetime
+import re
+
+from apistar.exceptions import ValidationError
+
+DATE_REGEX = re.compile(
```
audiolion: is the idea you subclass apistar.formats.DateFormat and provide your own DATE_REGEX if you want to support multiple or different formats, e.g. `2017/1/20`?

tomchristie: I've not gotten as far as making the formats configurable yet. Something along those lines will need to come, yup.

Having said that, in the particular case of date/time/datetime we might only support ISO formats in the core library, and force developers to use a custom `format` class if they want some other datetime behavior.

tomchristie:
Some notes on how this is coming along, here... https://discuss.apistar.org/t/a-peek-at-the-upcoming-version/412

apistar/server/app.py
```diff
@@ -16,37 +16,35 @@ def exception_handler(exc: Exception) -> Response:
...
+        return Router(document)
+
+    def init_templates(self, template_dir: str=None, template_apps: list=None):
+        return Templates(template_dir, template_apps, {
+            'reverse_url': self.reverse_url,
+            'static_url': self.static_url
+        })
+
+    def init_injector(self, components=None):
+        components = [] if (components is None) else list(components)
```
audiolion: why the parens around `(components is None)`?

also, small nit, on line 20 you use the style
```python
components = list(components) if components else []
```
why not normalize on one style and use in both places?

tomchristie: Good call - I've gone with the simpler style.

> why the parens around (components is None)

I think `... if ... else ...` is easier to visually parse if any expression in the middle section is demarcated by being enclosed in parentheses. However we don't need that for single-variable expressions, so this is better anyways.

apistar/types.py
```diff
@@ -1,69 +1,100 @@
...
+class TypeMetaclass(ABCMeta):
+    def __new__(cls, name, bases, attrs):
+        properties = []
+        for key, value in list(attrs.items()):
+            if key in ['keys', 'items', 'values', 'get', 'validator']:
+                msg = (
+                    'Cannot use reserved name "%s" on Type "%s", as it '
+                    'clashes with the class interface.'
+                )
+                raise ConfigurationError(msg % (key, name))
```
GomZik: Maybe better to move properties somewhere in cls._properties to exclude name conflicts?

apistar/compat.py
sirex: > The order-preserving aspect of this new implementation is considered an implementation detail and should not be relied upon

https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-compactdict

I could not find anything about guaranteed dict order since 3.7+.

sirex:
Where is the third party async for apistar?

sirex:
I think, if WSGI is supported, then ASGI should be supported too.

audiolion:
its called uvicorn and uvitools, Tom just put them in a separate repo

sirex:
@audiolion neither uvicorn/uvitools, nor apistar 4.0 depends on each other, so it looks, that there is no third party library that would bring ASGI support to apistar 4.0.

If ASGI is wiped out of apistar, why WSGI is an exception?

tomchristie:
@sirex - 0.4 Is a major rewrite - I've tackled WSGI first, because that's the simpler case. Interestingly ASGI's pretty different since 0.3 was released.

At the time it didn't support the sort of interface we needed for APIStar, because it was designed around how it's used with Django and channels, so instead we were using "an ASGI-based" interface -
 the [Uvicorn Messaging Interface].(https://github.com/encode/uvicorn/blob/master/docs/index.md#messaging-interface))
0.4 might be the right time to see if we can fall into line with ASGI now. (It'd be pretty great if so)

apistar/compat.py
jgirardet: The boss said it : https://mail.python.org/pipermail/python-dev/2017-December/151283.html

skion:
I like how you synchronised the PR number with the release version for the occasion.

tomchristie:
@skion Me too. 🍾😎

pikeas:
Is there a best guess for when .4 will land? Trying to schedule updating one of our apps into our sprint planning cycle.

jgirardet:
hello @tomchristie, 
did you ever consider using cerberus at http://python-cerberus.org/ ? I just discovered it and It's finally the same thing you're trying to do with types and validators

jgirardet:
I take a look  at cerberus and It works very easy with apistar with a small component.

I put it there : https://gist.github.com/jgirardet/8b666c949ba547121598b90777118bf5 if somebody wants to see. Could do a good third party app.

Furthemore, this shows how component system is much better and very easy to use :heart_eyes:

tomchristie:
Love that example! 😀

sirex:
@tomchristie tests for ASGI https://github.com/encode/apistar/pull/414

jgirardet:
another try subclassing cerberus's validator : https://gist.github.com/jgirardet/8b666c949ba547121598b90777118bf5 It's very easy to use and highly customisable.
 It really  fits my need, I think I will continue in this way. I 'll give you some news later  on discuss

jordanambra:
@tomchristie are there issues that you could reference in that TODO list? Some of those items are pretty opaque for trying to understand from an outsider's perspective. Maybe some of us could help provide implementation if you wanted, but aren't sure what those TODOs reference or if there's a spec or description of what you're trying to do.

Looking forward to 0.4! 0.3 on Zappa has been serving me very well for about half a year now.

README.md
```diff
@@ -504,14 +504,144 @@ Validates list input.
...
+```
+
+**app.py**:
+
+```
+from apistar import App, Route
+
+
+def welcome(app: App, name=None):
+    return app.render_template('index.html', name=name)
```
sirex: Is there a possibility to render same handler using different renderers specified in route or accept header?

For example:


```python
def welcome(name: str=None) -> dict:
    return {'name': name}

routes = [
    Route('/', method='GET', handler=welcome, renderer=template('index.html')),
    Route('/', method='GET', handler=welcome, renderer='json'),
    Route('/', method='GET', handler=welcome, renderer='yaml'),
]
```

Or renderer could be chosen by accept header:

```python
routes = [
    Route('/', method='GET', handler=welcome, renderer={
        'text/html': template('index.html'),
        'application/json': 'json',
        'application/x-yaml': 'yaml',
    }),
]
```

README.md
```diff
@@ -504,14 +504,144 @@ Validates list input.
...
+
+app = App(routes=routes)
+
+
+if __name__ == '__main__':
+    app.serve('127.0.0.1', 8080, use_debugger=True)
+```
+
+You can configure where templates are served from by using the `template_dir`
+argument when instantiating an application.
```
sirex: As I understand `installed_packages` also can be used to control template search paths?

thedrow:
FYI the [`injector`](https://github.com/alecthomas/injector) package dropped support for Python 2.7 and now supports the same style of annotations we use for dependency injection.
I'd recommend moving away from a custom implementation as the design and implementation of `injector` is, in my opinion, more complete and more mature.

jgirardet:
```
    def serve(self, host, port, use_debugger=False, **options): 
            werkzeug.run_simple(host, port, self, **options)
```
I saw merged my pr  it but it's not yet in the app.py file (ASGI is ok but not  App)

it's still
```
def serve(self, host, port, use_debugger=False):
   werkzeug.run_simple(host, port, self, use_debugger=use_debugger)
```

jgirardet:
```
    def serve(self, host, port, use_debugger=False, **options): 
            werkzeug.run_simple(host, port, self, **options)
```
I saw merged my pr  it but it's not yet in the app.py file (ASGI is ok but not  App)

it's still
```
def serve(self, host, port, use_debugger=False):
   werkzeug.run_simple(host, port, self, use_debugger=use_debugger)
```

tomchristie:
@jgirardet - That's resolved now. Thanks!

jgirardet:
In case of 404 requests, the on_request hook is not called but the on_error is called. The on_error has no idea if the on_request things were done or not. 
In my case I had failure because on_errors had to close a database transactino which wasn't opened.
I don't know if it's a big problem or not but I think that user should be aware of that special case to deal with it because  404 seems to be the only situation where it happens (I tink because url lookup occurs before injection)

apistar/server/app.py
```diff
@@ -0,0 +1,274 @@
...
+    def init_hooks(self, event_hooks=None):
+        if event_hooks is None:
+            event_hooks = []
+
+        self.on_request_functions = [
+            hook.on_request for hook in event_hooks
+            if hasattr(hook, 'on_request')
+        ]
+
+        self.on_response_functions = [self.render_response] + [
```
Bogdanp: I think this list and the `on_error` list should be reversed so that request and response hooks are always at the same depth in the request stack. I went into more details why [here](https://discuss.apistar.org/t/homing-in-on-0-4/426/9?u=bogdanp).

PR #405, opened 2018-03-21
Exception returns dict with key "msg"

jgirardet:
it's more consistent with the rest of the api


tomchristie:
Probably will do something like this at some point yeah, eg making sure plain strings end up as {“error”: ...}

PR #406, opened 2018-03-21
None or '' raise Attribut Error : object has no attribute iso format

jgirardet:


audiolion:
I feel like `to_string()` should return `''` instead of `None`

jgirardet:
Make sens :stuck_out_tongue:

tomchristie:
More generally what we want to make sure is that `DateTimeValidator` etc... properly support `allow_null`. I think we probably *should* serialize unset date/times as `None` rather than ''. Not sure if that implies we should rename `to_string`, since the `None` case is always going to be there for nullable validators.

jgirardet:
to_string is meaningful.
to_serializable is not very nice ;-)
Not a problem if It's said in the doc that null date is None

tomchristie:
Thanks folks! Closed via https://github.com/encode/apistar/commit/e8f60f6104e0ad49cf365d0b5941ad15c737128c
Note that `Type` handles the `None` cases so we don't need to deal with it in each Validator.

PR #407, opened 2018-03-22
item without format fails in getitem

jgirardet:


tomchristie:
Thanks! Closed via https://github.com/encode/apistar/commit/e8f60f6104e0ad49cf365d0b5941ad15c737128c

PR #408, opened 2018-03-22
Email Validator

jgirardet:
re.compile type is tricky, you get it at runtime or via the private object. 

I added a file test_validator because I didn't know where to put it.

Maybe A new file called "addon_validator.py" separated from "base validators" could be a good thing.

the regex is stolen at django

### DIFF
--- a/apistar/validators.py
+++ b/apistar/validators.py
@@ -632,6 +632,7 @@ class Uniqueness():
 
         return element
 
+
 class Email(String):
     def __init__(self, **kwargs):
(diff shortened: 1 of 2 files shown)


jgirardet:
and the regex is stolen at django

tomchristie:
We will want `email`, `url` and possibly other formats too, but I'll need to think about this a bit first.

tomchristie:
Welcome to punt something similar into https://github.com/encode/typesystem, though I think we ought to just validate against something simple like `"[^@]+@[^@]+"`

PR #410, opened 2018-03-24
Required field

jgirardet:
Background  : a field required was removed in this new version. Now we have to specify "default" fo each non required field. This make the Types declaration very verbose. For example If you have a "create" schema, you only want a few required field but for most they aren't required so we have to write many things to achieve this.
The default parameter is not the same thing for all kind of validators. for String you can specifiy ''. For integer it's None and we have to add allow_null.
There is also a problem with the "enum" parameter. you can't specify a default (None or '') if it's not in the enum list or tuple.

For all those things a separated "required field" is much easier to use. Therefore it's more explicit (zen of python blabla...) and it's a bit faster (ok only 0,000000000001s) because you don't loop over properties.

In my PR I propose to keep both possibilities (via default or via field) even I think the default parameter way is not a good one.

Last thing : If there is a way to inherite a Type class (I really like to knno if it's possible and how), it will be much more easier with a separated "required" field.

### DIFF
--- a/apistar/types.py
+++ b/apistar/types.py
@@ -1,5 +1,5 @@
 from abc import ABCMeta
-from collections.abc import Mapping
+from collections.abc import Mapping, Sequence
 
 from apistar import validators
 from apistar.exceptions import ConfigurationError, ValidationError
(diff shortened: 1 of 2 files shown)


tomchristie:
Thanks for following everything that's going on!

I don't want "required" at the moment. That ends up giving you too many different ways around of saying "no value". For example strings could either be `""`, `null`, or `not present`. Using default values for "is this required" is more constrained and means we have Type objects that always have a fixed set of fields. For anything more flexible there's the generic `Object()` for open dictionary types.

> If there is a way to inherite a Type clas

Currently in #400, no. But there will be eventually, yes.

PR #411, opened 2018-03-26
permit werkzeug run custom option

jgirardet:
maybe, use_reloader=True should also be a default option for development.


apistar/server/app.py
```diff
@@ -69,8 +69,8 @@ def render_response(self, data):
             return HTMLResponse(data)
         return JSONResponse(data)
 
-    def serve(self, host, port, use_debugger=False):
-        werkzeug.run_simple(host, port, self, use_debugger=use_debugger)
+    def serve(self, host, port, use_debugger=False, **kwargs):
```
tomchristie: I guess let's drop `use_debugger=False`, and just pass `**options` through for everything.

PR #412, opened 2018-03-27
bugfix run_after_handler and run_on_exception

jgirardet:
I don't know if it's a bug or implementation choice, but I can't make it work without those changes.


tomchristie:
You'd pretty much always want to include those two yup!

I'm going to rejig these into a single `event_hooks` parameter, which takes a list of objects that can have `on_request`, `on_response` and `on_exception` hooks, so I'll close this off for now given that it's still got a bit of reworking to do.

PR #414, opened 2018-03-31
Tests for ASGI interface

sirex:
Added ASGI adapter for test client and updated http tests to use parametrized fixture, now all tests using `client` fixture will be called twice with WSGI and ASGI apps.

Tests helped to find these issues:

- QueryStringComponent didn't decoded bytes to str.

- HeadersComponent didn't include 'host' header.

- BodyComponent didn't returned body at all. Current body retrieval does not support partial bodies, probably this should be added later.

### DIFF
--- a/apistar/server/app.py
+++ b/apistar/server/app.py
@@ -17,6 +17,8 @@ from apistar.server.wsgi import WSGI_COMPONENTS, WSGIEnviron, WSGIStartResponse
 
 
 class App():
+    interface = 'wsgi'
+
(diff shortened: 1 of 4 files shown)


tomchristie:
Lovely stuff, thank you! Good to see how we ought to be using py.test fixtures properly there.

Coupla tweaks to be made. I'll just pull this in and then deal with them.

* The host header should be added by the test client rather than being automatically included.
* Reading the full request body.

Issue #130, opened 2017-04-29
pypy3 support


piotr-szpetkowski:
In theory I believe that latest pypy (that is v5.7.1) should support all of the features required by api* as it implements Python 3.5.3. Most dependencies are already supported out of the box - even some of those written in C (http://packages.pypy.org/##ujson).

I belive there are performance benefits in using pypy3 instead of using CPython - even for APIs. 
So - is there a reason to not to include pypy3 as supported?

tomchristie:
Would be more than happy to do so. Does Travis support pypy3?

mammuth:
There is an open issue at the Travis repo for pypy3 5.7 support (https://github.com/travis-ci/travis-ci/issues/7507). 
A comment in the issue apparently shows a working solution/workaround for pypy 5.7 (https://github.com/open-craft-guild/aiohttp-ripozo/blob/4f1c963b59387a3637fa31fda4500d5d604a9520/.travis.yml#L10-L61)

piotr-szpetkowski:
Would using docker (https://docs.travis-ci.com/user/docker/) work?

tomchristie:
I don't really want to make any great changes to our CI in order to accommodate pypy3 at this point.
As a first pass I'd suggest installing and running the tests locally under pypy3 and reporting back on this thread.

tomchristie:
Following on from that I'd then suggest that I'd document the benchmarking, so that you can test it under pypy3 for comparison.

piotr-szpetkowski:
pypy3 with git master has following result:

> 98 passed in 1.77 seconds

tomchristie:
I'm going to close this off for now. If the tests pass then, great. We're not able to do anything low effort in order to test it in our CI, tho, so there's nothing actionable for us to do here right now. I'd be perfectly happy for someone to add pypy3 to our techempower benchmarks if it's available, and I'd be happy for us to add pypy3 on Travis once a 3.5+ compatible version is available there.

sanxiyn:
FYI, PyPy3.5 is now available on Travis.

tomchristie:
Gotcha. Pull requests adding it to the matrix would be accepted then.

PR #418, opened 2018-04-06
Add PyPy3.5 to the testing matrix

necaris:
Should help finish off #130

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ cache: pip
 python:
     - "3.5"
     - "3.6"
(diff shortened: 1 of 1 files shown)


tomchristie:
Thanks! I've pulled this into the 0.4 branch. (See #400)

necaris:
Thanks @tomchristie ! I'm seeing some test failures because `mypy`'s dependency `typed-ast` doesn't play well with PyPy -- looking into travis config to fix that so we can actually get tests running...

PR #421, opened 2018-04-06
Only require mypy if not running under PyPy

necaris:
Since its dependency `typed_ast` doesn't play well outside of CPython. Uses environment markers in the requirements file (https://www.python.org/dev/peps/pep-0508/#environment-markers) to exclude the PyPy environment.

Includes fixes to the Travis-CI configuration to make sure tests continue to run in all environments.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,5 +11,6 @@ whitenoise
 codecov
 flake8
 isort
+# mypy ; implementation_name != 'pypy'
 pytest
 pytest-cov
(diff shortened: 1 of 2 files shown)


tomchristie:
Doesn’t look like the mypy branch is running...

See the test outputs, eg:

+[[ 3.6 != pypy* ]]
scripts/lint: 12: scripts/lint: [[: not found

necaris:
@tomchristie apologies for the delay, fixed the expression to work in Ubuntu `dash` and disabled it by default.

tomchristie:
Given that it's all commented out at the moment, I'd rather close this off, and only reopen if we decided to bring mypy testing back in.

PR #423, opened 2018-04-09
when HTTP404, only on_error occurs.

jgirardet:


tomchristie:
So I think this is maybe something that needs to change. Might well make sense for `on_error` to only be called for unhandled cases, but have the `APIException` subclasses result in the usual `on_response` behavior.

PR #424, opened 2018-04-11
Fix example in type system documentation

mammuth:
When reading the example in _Type System > Formats_ it looks like some copy/paste mistakes happened here.

This addresses the new documentation in 0.4


mammuth:
Merged in order to get the build running again (https://github.com/encode/apistar/commit/ce317862fce3ffc92126e1cbcb21024b9890f5ad)

Issue #427, opened 2018-04-11
No form encoding support in 0.4


Bogdanp:
It looks like [the only](https://github.com/encode/apistar/blob/3aed0e4b90c6c949b2a915b8e750658d87b0f7d5/apistar/server/validation.py#L16) codec that `RequestDataComponent` supports in 0.4 is `JSONCodec`. Are you planning to add support for form posts to `RequestData` later or are apps expected to roll their own `FormData` (or similar) component?

tomchristie:
> Are you planning to add support for form posts to RequestData

Yup.

tomchristie:
That’s a good easy one that ought to sit towards the top of our TODO pile.

tomchristie:
I think we’d start with hardcoded json+form parsing, and treat configurable parsers/renderers as a later enhancement.

Bogdanp:
Sounds good to me!

tomchristie:
Closed via #430.

PR #430, opened 2018-04-12
Add form data codecs

Bogdanp:
Per discussion in #427, I figured I'd jump in and help out.

### DIFF
--- a/apistar/codecs/formdata.py
+++ b/apistar/codecs/formdata.py
@@ -1,0 +1,36 @@
+from io import BytesIO
+from itertools import chain
+
+from werkzeug.datastructures import ImmutableMultiDict
+from werkzeug.formparser import FormDataParser
+from werkzeug.http import parse_options_header
+from werkzeug.urls import url_decode
+
+from apistar.codecs.base import BaseCodec
+
+
+class MultiPartCodec(BaseCodec):
+    media_type = 'multipart/form-data'
+
+    def decode(self, bytestring, headers, **options):
+        try:
+            content_length = max(0, int(headers['content-length']))
+        except (KeyError, ValueError, TypeError):
+            content_length = None
+
+        try:
+            mime_type, mime_options = parse_options_header(headers['content-type'])
+        except KeyError:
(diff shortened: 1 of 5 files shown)


apistar/codecs/formdata.py
```diff
@@ -0,0 +1,34 @@
...
+            content_length = max(0, int(headers['content-length']))
+        except (KeyError, ValueError, TypeError):
+            content_length = None
+
+        try:
+            mime_type, mime_options = parse_options_header(headers['content-type'])
+        except KeyError:
+            mime_type, mime_options = '', {}
+
+        body_file = BytesIO(bytestring)
```
Bogdanp: This should be somewhat of a stopgap since this codec should have access to a file-like object rather than the framework having to read the whole request body into memory, but I couldn't figure out a quick way to get to the underlying file object. Let me know if/how I should change this.

audiolion: You could take inspiration from how `flask-api` does it:

https://github.com/flask-api/flask-api/blob/376815deca2c6ec827633e9e152e7452de1540dd/flask_api/parsers.py#L33-L57

Bogdanp: I would, but the problem is getting the `stream` (body file) parameter into `decode(...)` in the first place. [RequestDataComponent](https://github.com/encode/apistar/blob/3aed0e4b90c6c949b2a915b8e750658d87b0f7d5/apistar/server/validation.py#L14) depends on `Body` and I believe that represents the full request body as a bytestring. There doesn't seem to be a `BodyFile` or similar component so I figured I'd ask before going deeper. :D

Bogdanp:
Is there anything in particular holding this up? It doesn't seem possible to work around the issue in-app since `RequestDataComponent` attempts to parse the request body automatically so, even if I register my own `FormDataComponent`, I can't get this to work. I'll happily drop the `MultiPartCodec` if that's what's holding this up, since I don't need it.

tomchristie:
Awesome stuff!

PR #439, opened 2018-04-14
Allow validators as params annotations

Aulos:
For some params it would be nice to have better validators, ex. enums or patterns.


Aulos:
It should be possible to validate params:

```python
def handler(path_param: validators.String(enum=['val1', 'val2']), query_param: validators.String(max_length=5)):
    return {"result": "OK"}
```

or

```python
class MyValidator(validators.String):
    def __init__(self, **kwargs):
        kwargs.pop('min_length', None)
        super().__init__(min_length=10, **kwargs)

def handler(param: MyValidator()):
    return {"result": "OK"}
```

audiolion:
```python
class Title(types.Type):
    title = validators.String(min_length=10)

def handler(param: Title):
    return {"result": "OK"}
```

what is wrong with using types?

Aulos:
`Type` indicates an object and it's always from body. Currently you cannot validate query or path params.

audiolion:
ah I understand the use case now. I think it would be better to validate route params at the `Route` level imo, otherwise you are duplicating the logic for validating a UUID across every function that uses it *and* it has to go farther down the stack before realizing the request is invalid.

Aulos:
That's exactly what I'm proposing here.

audiolion:
Ah right, I forgot that Routes only allow one handler/method

Aulos:
To get headers you also use `http.Header` and to get query param you can use `http.QueryParam` (`str` works better). So using primitive types for path/query params seems like unreasonable limitation - why not say that you should use primitive types, but you *can* use validator if you need to?

tomchristie:
True, tho at least in our codebase, we really *do* pass those types. eg...

https://github.com/encode/apistar/blob/master/apistar/server/wsgi.py#L83
https://github.com/encode/apistar/blob/master/apistar/http.py#L13

But yes, we don't particular expect or require users to do the same.

For the moment I'd probably like us to avoid this, but add support for `Type` on query and path parameters. A third party component for injecting validator-annotated parameters would be perfectly doable, right?

Aulos:
Yup, but it won't appear in the schema then :-|

Aulos:
Btw. using `http.QueryParam` also causes these params to not show up in schema - only using primitives does (https://github.com/encode/apistar/blob/master/apistar/server/core.py#L45)

PR #440, opened 2018-04-14
Make minor cleanup and improve readability

imanhodjaev:
This small PR includes the following changes

* Use set literal instead of initiating set,
* Make `__str__` in `compat.py` reachable,
* Remove class definitions with empty parens.

### DIFF
--- a/apistar/compat.py
+++ b/apistar/compat.py
@@ -38,8 +38,8 @@ try:
             mode = "" if self.closed else " '%s'" % self.file.mode
             return "<DownloadedFile '%s', %s%s>" % (self.name, state, mode)
 
-            def __str__(self):
-                return self.__repr__()
+        def __str__(self):
+            return self.__repr__()
 
 except ImportError:
     # On some platforms (eg GAE) the private _TemporaryFileWrapper may not be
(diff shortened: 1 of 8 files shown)


apistar/compat.py
```diff
@@ -38,8 +38,8 @@ def __repr__(self):
             mode = "" if self.closed else " '%s'" % self.file.mode
             return "<DownloadedFile '%s', %s%s>" % (self.name, state, mode)
 
-            def __str__(self):
-                return self.__repr__()
+        def __str__(self):
+            return self.__repr__()
```
Bogdanp: Why not `repr(self)`?

imanhodjaev: I just wanted to keep it as it is, so in next PRs I or someone else can do it.

tomchristie:
Great, yup!

tomchristie:
One thing we could consider for formatting aspects would be using ‘black’ to enforce a common style throughout. I don’t know how that looks in practice, or if it’s sonething we definitely want, but might be worth is taking a look at.

imanhodjaev:
Great then I will try to configure `black` and add formatting script for it.

PR #442, opened 2018-04-15
Apply automatic code formatting using black

imanhodjaev:
This PR is to see if `black` code formatting works well for auto formatting.
This might break `isort` import order checks

```
(apistar-YBJ7r3xm) 11:22 ≡ scripts/lint                                                                                                    ᴦ/P/ apistar ≡ REFACTOR/CODE-FORMAT
+ flake8 apistar tests
+ isort apistar tests --recursive --check-only
ERROR: ~/Projects/apistar/apistar/server/app.py Imports are incorrectly sorted.
ERROR: ~/Projects/apistar/tests/test_client.py Imports are incorrectly sorted.
```

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,6 +1,6 @@
 [settings]
 atomic=true
-multi_line_output=5
+multi_line_output=3
 not_skip=__init__.py
 known_standard_library=types
 known_third_party=pytest
--- a/apistar/__init__.py
+++ b/apistar/__init__.py
@@ -11,8 +11,18 @@ from apistar.document import Document, Field, Link, Section
 from apistar.server import App, ASyncApp, Component, Include, Route
 from apistar.test import TestClient
 
-__version__ = '0.4.3'
+__version__ = "0.4.3"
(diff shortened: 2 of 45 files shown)


audiolion:
If we modify (or add) and `.isort.cfg` or editor config file and use the `3 - Vertical Hanging Indent` option for imports it should be ok

https://isort.readthedocs.io/en/latest/#configuring-isort

imanhodjaev:
@audiolion done! 🍰

tomchristie:
Okay, so I think a good call here would be to first make a PR that introduces `black` into the dev requirements and continuous integration, but doesn't apply the changes that it makes. The PR should fail at first, demonstrating that when code hasn't been formatted it doesn't pass the style check. We can then add a second commit, which is the result of applying black to the codebase.

imanhodjaev:
@tomchristie sorry for delay, and yes I am completely agree with you, so first I will introduce black under `scripts/format` or `scripts/black` and then we see if it works out well. Thanks for review :)

Issue #435, opened 2018-04-13
Type inheritance in 0.4.3


rhelms:
Is Type inheritance supported in 0.4.3?

For example, I have these two classes:

```
class NbnOrderFeasibilityRequest(types.Type):
    sqId = validators.String()
    nbnLocationId = validators.String()
    tc4DownloadSpeed = validators.Integer()
    tc4UploadSpeed = validators.Integer()
    cpiId = validators.String(default=None)
    ntdId = validators.String(default=None)
    potsInterconnectId = validators.String(default=None)
    transferType = NbnOrderTransferType(default=None)
    localNumberPorting = validators.String(enum=['Yes', 'No'], default=None)
    customerAuthorityDate = validators.Date(default=None),
    appointmentId = validators.String(default=None)
    installationWorkforce = NbnOrderInstallationWorkforceType(default=None)  # TODO conditions
    installationCentralSplitter = validators.String(enum=['Yes', 'No'], default=None)
    exchangePairConnection = validators.String(enum=['No'], default=None)
    ntdPortId = validators.Integer(default=None)
    batteryBackup = validators.String(enum=['Yes', 'No'], default=None)

class NbnOrderRequest(NbnOrderFeasibilityRequest):
    accessSeekerRef = AccessSeekerRef(default=None)
```

When I instantiate an NbnOrderRequest, only accessSeekerRef seems to be available.

    data = {
        'sqId': 'SQ00001',
        'nbnLocationId': 'LOC00001',
        'tc4DownloadSpeed': 10,
        'tc4UploadSpeed': 10
    }
    request = NbnOrderRequest(data)
    assert {'accessSeekerRef': None) == dict(data) # Works, and it shouldn't

I'm guessing there's something that needs to happen in the __new__ operation of TypeMetaclass to get fields/attributes from the parent class, but I just wanted to check if this is supposed to be possible.

tomchristie:
It’s not supported yet, would need a little work on the metaclass. Yes I’d like to see it added. If anyone wants to take it on it’d be worth looking at the metaclass implementation of Django REST framework ModelSerializer

tomchristie:
Also worth noting. In JSON schema we’ll want to map these to allOf cases, so we can included the child and parent definitions. (Tho only works in the case of extending fields, not overriding them)

rhelms:
I'm having a crack at this now.
I've duplicated most tests in test_types.py to make sure the expected operations work with an extended Type.
[...]

PR #444, opened 2018-04-16
Allow Type extension

rhelms:
Does not implement allOf in JSON Schema
encode/apistar#435

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -151,59 +151,6 @@ def test_setattr():
     assert exc.value.detail == 'May not be null.'
 
 
-def test_extended_setattr():
-    when = datetime.datetime(2020, 1, 1, 12, 0, 0, tzinfo=utc)
-    product = ReviewedProduct(name='abc', rating=20, created=when, reviewer='me')
-
-    assert product.name == 'abc'
-    assert product.rating == 20
-    assert product.created == when
-    assert product.reviewer == 'me'
-    assert dict(product) == {
-        'name': 'abc',
-        'rating': 20,
-        'created': '2020-01-01T12:00:00Z',
-        'reviewer': 'me'
-    }
-
-    # Set a format field, using a native type.
-    product.created = datetime.datetime(2030, 2, 2, 13, 30, 0, tzinfo=utc)
-    assert dict(product) == {
-        'name': 'abc',
-        'rating': 20,
-        'created': '2030-02-02T13:30:00Z',
-        'reviewer': 'me'
-    }
-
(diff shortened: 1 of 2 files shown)


tomchristie:
Great stuff. I'd prefer that we don't introduce so much extra code into the test cases, as I think it's unnecessarily verbose there. I'll comment on those cases inline...

tests/test_types.py
```diff
@@ -35,6 +39,23 @@ def test_from_object():
     }
 
 
+def test_extended_from_object():
```
tomchristie: Yup this is reasonable.

tests/test_types.py
```diff
@@ -49,6 +70,22 @@ def test_from_kwargs():
     }
 
 
+def test_extended_from_kwargs():
```
tomchristie: Let's have this one too.

tests/test_types.py
```diff
@@ -63,6 +100,22 @@ def test_from_dict():
     }
 
 
+def test_extended_from_dict():
```
tomchristie: Okay, for consistency.

tests/test_types.py
```diff
@@ -98,6 +151,59 @@ def test_setattr():
     assert exc.value.detail == 'May not be null.'
 
 
+def test_extended_setattr():
```
tomchristie: I think this becomes unnecessary.

tests/test_types.py
```diff
@@ -133,6 +239,59 @@ def test_setitem():
     assert exc.value.detail == 'May not be null.'
 
 
+def test_extended_setitem():
```
tomchristie: Again, let's drop this one.

tests/test_types.py
```diff
@@ -150,6 +309,23 @@ def test_misc():
         Product([])
 
 
+def test_extended_misc():
```
tomchristie: Let's drop this too. Test cases are a maintenance cost too.

tests/test_types.py
```diff
@@ -187,3 +363,90 @@ def test_as_jsonschema():
...
+                    }
+                },
+                "required": [
+                    "reviewer"
+                ]
+            }
+        }
+    }
+
+    """
```
tomchristie: I'd suggest we drop the docstring here. The behavior's currently good as it is, we don't need to document what we may intend to do in the future.

rhelms:
Review changes have been committed.

tomchristie:
Excellent, thank you!

PR #449, opened 2018-04-17
Fix _WSGIAdapter and TestClient WSGI compliance

pslacerda:
The missing `exc_info=None` parameter was causing me throuble when playing with middlewares and `TestClient`. This fix shoud be straightforward.

### DIFF
--- a/apistar/test.py
+++ b/apistar/test.py
@@ -183,7 +183,8 @@ class _ASGIAdapter(requests.adapters.HTTPAdapter):
 class _TestClient(requests.Session):
     def __init__(self, app: typing.Callable, scheme: str, hostname: str) -> None:
         super(_TestClient, self).__init__()
-        if app.interface == 'asgi':
+        interface = getattr(app, 'interface', None)
+        if interface == 'asgi':
             adapter = _ASGIAdapter(app)
         else:
             adapter = _WSGIAdapter(app)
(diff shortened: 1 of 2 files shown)


apistar/test.py
```diff
@@ -77,7 +77,9 @@ def send(self, request, *args, **kwargs):
         """
         raw_kwargs = {}
 
-        def start_response(wsgi_status, wsgi_headers):
+        def start_response(wsgi_status, wsgi_headers, exc_info=None):
+            if exc_info:
+                raise exc_info[0].with_traceback(exc_info[1], exc_info[2])
```
tomchristie: Should this instead be `raise exc_info[0](exc_info[1]).with_traceback(exc_info[2])`?

Eg. see https://docs.python.org/3/library/exceptions.html#BaseException.with_traceback and https://docs.python.org/3.6/library/sys.html#sys.exc_info

tomchristie: Also we *might* want to consider a test case.

pslacerda:
For sure I use TestClient to have an batch API that emulate many requests when the client did just one. But I didn't expected to test the test framework.

I make lots of sense however.

pslacerda:
If you already reviewed it, sorry for this ammend. My linter isn't working correctly.

PR #455, opened 2018-04-17
Add handler docstring to OpenAPI's description

pslacerda:
Both description and summary was missing at OpenAPI generated schema. This set the path description to the handler docstring. However I found no suitable text for summary.


tomchristie:
Yup, looks good!

PR #460, opened 2018-04-18
Exception handling

### LINKED ISSUE #219: could it return {"error": "exception "} instead of entering the debugger?
e.g.:

`{"error": "ValueError: invalid literal for int() with base 10: 'a'"}`

At least this is what I would expect if not in debugging mode...

### LINKED ISSUE #437: Error hooks don't fire on uncaught exceptions
In 0.4 right now, if a handler raises an exception that's not a standard API Star error, then any registered `on_error` hooks don't fire. Minimal example:

```python
from apistar import App, Route, http, exceptions


def index() -> dict:
    return {}


def fail_1() -> None:
    raise Exception("fail")


def fail_2() -> None:
    raise exceptions.BadRequest("bad request")


class Hook:
    def on_error(self, response: http.Response) -> http.Response:
        print("ON ERROR HOOK WAS CALLED")
        return response


app = App(
    routes=[
        Route("/", "GET", index),
        Route("/fail-1", "GET", fail_1),
        Route("/fail-2", "GET", fail_2),
    ],
    event_hooks=[Hook()]
)


app.serve(host="127.1", port=8000)
```

When you visit `/fail-2` the message gets printed to stdout, but when you visit `/fail-1`, it doesn't.

### LINKED ISSUE #443: start_response not being called disrespecting WSGI
When an exception other than `HTTPException` was in the handler `App.finalize_wsgi()` isn't called, disrespecting the WSGI standard.

```python
def welcome():
    raise Exception
```

tomchristie:
Closes #437.
Closes #443.
Closes #449.
Closes #423.
Closes #219.

* `on_response` now runs for both regular reponses and handled exceptions.
* `on_error` now only runs for unhandled exceptions (ie. 500 reponses).
* Return 500 responses on error.
* Test client raises exceptions instead of returning 500 responses.
* Add `debug` flag to `app.serve()`, to switch between debugger vs. plain responses.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ app = App(routes=routes)
 
 
 if __name__ == '__main__':
-    app.serve('127.0.0.1', 5000, use_debugger=True, use_reloader=True)
+    app.serve('127.0.0.1', 5000, debug=True)
 ```
 
 ---
--- a/apistar/__init__.py
+++ b/apistar/__init__.py
@@ -11,7 +11,7 @@ from apistar.document import Document, Field, Link, Section
 from apistar.server import App, ASyncApp, Component, Include, Route
 from apistar.test import TestClient
 
-__version__ = '0.4.4'
+__version__ = '0.4.5'
 __all__ = [
     'App', 'ASyncApp', 'Client', 'Component', 'Document', 'Section', 'Link', 'Field',
     'Route', 'Include', 'TestClient', 'http'
--- a/apistar/http.py
+++ b/apistar/http.py
@@ -187,11 +187,13 @@ class Response:
     def __init__(self,
                  content: typing.Any,
                  status_code: int=200,
-                 headers: typing.Union[StrMapping, StrPairs]=None) -> None:
+                 headers: typing.Union[StrMapping, StrPairs]=None,
+                 exc_info=None) -> None:
         self.content = self.render(content)
         self.status_code = status_code
         self.headers = MutableHeaders(headers)
         self.set_default_headers()
+        self.exc_info = exc_info
 
     def render(self, content: typing.Any) -> bytes:
         if isinstance(content, bytes):
--- a/apistar/server/adapters.py
+++ b/apistar/server/adapters.py
@@ -10,8 +10,9 @@ class ASGItoWSGIAdapter(object):
     We want this so that we can use the Werkzeug development server and
     debugger together with an ASGI application.
     """
-    def __init__(self, asgi):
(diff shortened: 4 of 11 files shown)


pslacerda:
You forget the interface attribute used by TestClient. You need it to be optional for use TestClient with middlewares.

tomchristie:
@pslacerda I'm not sure what you mean.

We have this... https://github.com/encode/apistar/pull/460/files#diff-d5d81d6c1db21ec27d43e6695624620eR82 which ensures the TestClient raises if `exc_info` is included.

That behavior gets tested by https://github.com/encode/apistar/pull/460/files#diff-f7c09dd1e37b8dd75222f9a8a328dd6fR44 which ensures that the exception is raised by the TestClient (even though debug=False on the application)

pslacerda:
Also if any on_error hook raise an exception finalize_wsgi will not be called. Probably it should be an statement by itself at the end of run(), so it will only fails if a response couln't be injected.

pslacerda:
@tomchristie I'm talking about https://github.com/encode/apistar/pull/449/files#diff-d5d81d6c1db21ec27d43e6695624620eR186.

If the TestClient app hasn't the interface non standard attribute it will fails.

tomchristie:
Included the interface check, expanded the docs on `on_error`.

tomchristie:
Right, assuming no-one objects I'm going for a minor version bump on this.

pslacerda:
I see what you mean. Think about some monitoring middleware like AppOptics. Because start_response isn't called then it will not end its trace resulting in wrong monitoring data. This mean that we can't monitor the app from an external layer because there is no guarantee that it will always continue the middleware hooks.

tomchristie:
@pslacerda Happy to consider further refinements on top of this. It's possible that we could have a last-ditch catch and return. I'd assume any robust monitoring middleware will deal with exception cases. We could talk this out on the discussion group, or if there's a concrete example of a third-party WSGI middleware that this wouldn't play well with then raise a new issue for it?

pslacerda:
You don't got it. Now apistar is literally broken and disrespecting WSGI. Yes, no wsgi middlware wil work if on_error fails, specifically appoptics in my case, now I must change inside the app to guaramtee the right behavour.  And aso I object about bump version because break semantic versioning.

tomchristie:
> And aso I object 

Right, which was why I asked. There's a bit of fuzziness given that we're in alpha, but I've no problem with making a 0.5 bump.

PR #464, opened 2018-04-18
Example to process handled/unhandled exceptions in event hooks

PeRDy:
Adds an example to show how to process handled and unhandled errors in event hooks:

```python
class ErrorHandlingHook():
    def on_response(self, response: http.Response, exc: Exception):
        if exc is None:
            print("Response OK")
        else:
            print("Handled exception")
            
    def on_error(self, response: http.Response):
        print("Unhandled error")

event_hooks = [ErrorHandlingHook()]

app = App(routes=routes, event_hooks=event_hooks)
```


tomchristie:
Thanks! I might do some tweaking in how we present this, but let's pull it in as-is to start with!

PR #482, opened 2018-04-24
WIP: Add Validated(Path|Query)Param annotation

Aulos:


pslacerda:
Forget the last comment.  Just for reference, I implemented custom path param validators with the following component. 

```python
class ValidatorParamComponent(Component):

    def can_handle_parameter(self, parameter: inspect.Parameter):
        return issubclass(parameter.annotation, validators.Validator)

    def resolve(self,
                parameter: inspect.Parameter,
                path_params: ValidatedPathParams,
                query_params: ValidatedQueryParams):
        params = path_params if (parameter.name in path_params) else query_params
        has_default = parameter.default is not parameter.empty

        validator = validators.Object(
            properties=[(parameter.name, parameter.annotation())],
            required=[] if has_default else [parameter.name]
        )

        try:
            params = validator.validate(params, allow_coerce=True)
        except validators.ValidationError as exc:
            raise exceptions.NotFound(exc.detail)
        return params.get(parameter.name, parameter.default)
```

PR #491, opened 2018-04-26
Fix coercion when issubclass receive an instance

pslacerda:
Schema responses wasn't being generated or was failing with the following examples.
 
```python

def handler() -> Number:
    # fails because of logic 
    pass

def handler() -> Number():
    # fails because issubclass receive an instance
    pass
```


pslacerda:
So I guess this PR should be adjusted to deal with validator classes and primitive types instead merged accepting also validator instances, no?

PR #493, opened 2018-04-26
API Docs

tomchristie:
Work in progress.

image


GomZik:
Hooray! Tom, please, take a look at how schema generates for Includes. Looks like Include path not used for paths key in schema.

GomZik:
Also, swagger ui trying to OPTIONS before POST. Can we have autogenerated OPTIONS method?

tomchristie:
screen shot 2018-05-08 at 12 55 39

tomchristie:
> take a look at how schema generates for Includes. Looks like Include path not used for paths key in schema.

Yup, resolved in this PR.

tomchristie:
Note that the API docs are not currently interactive.
We'll tackle that as an incremental improvement.

PR #494, opened 2018-04-26
Fix bug with section name

GomZik:


tomchristie:
The change you've included here *is* correct (so I'll pull it in as-is). However we oughta drop the `section=` and `link=` arguments completely. They've been left there incorrectly.

Issue #496, opened 2018-04-26
Array validation does not work


marekkosta:
Regard the following:
```python
from apistar import App, Route
from apistar import types, validators


class Numbers(types.Type):
    numbers = validators.Array(validators.Integer)


def test(nums: Numbers)-> dict:
    res = [x+1 for x in nums['numbers']]
    return {'numbers': res}


routes = [
    Route('/test', method='POST', handler=test),
]

app = App(routes=routes)


if __name__ == '__main__':
    app.serve('127.0.0.1', 8000, debug=True)
```
then I go:
```bash
curl -X POST -H 'Content-Type: application/json' -i http://localhost:8000/test --data '{ "numbers": [1,2,3] }'
```
This crashes with the following Traceback:
```bash
Traceback (most recent call last):
  File "/home/mkosta/.local/share/virtualenvs/apistar-test-b1c5t0T_/lib/python3.6/site-packages/apistar/server/app.py", line 222, in __call__
    return self.injector.run(funcs, state)
  File "/home/mkosta/.local/share/virtualenvs/apistar-test-b1c5t0T_/lib/python3.6/site-packages/apistar/server/injector.py", line 106, in run
    state[output_name] = func(**func_kwargs)
  File "/home/mkosta/.local/share/virtualenvs/apistar-test-b1c5t0T_/lib/python3.6/site-packages/apistar/server/app.py", line 166, in finalize_wsgi
    raise exc_info[0].with_traceback(exc_info[1], exc_info[2])
  File "/home/mkosta/.local/share/virtualenvs/apistar-test-b1c5t0T_/lib/python3.6/site-packages/apistar/server/app.py", line 215, in __call__
    return self.injector.run(funcs, state)
  File "/home/mkosta/.local/share/virtualenvs/apistar-test-b1c5t0T_/lib/python3.6/site-packages/apistar/server/injector.py", line 106, in run
    state[output_name] = func(**func_kwargs)
  File "/home/mkosta/.local/share/virtualenvs/apistar-test-b1c5t0T_/lib/python3.6/site-packages/apistar/server/app.py", line 206, in __call__
    return self.injector.run(funcs, state)
  File "/home/mkosta/.local/share/virtualenvs/apistar-test-b1c5t0T_/lib/python3.6/site-packages/apistar/server/injector.py", line 106, in run
    state[output_name] = func(**func_kwargs)
[...]

PR #497, opened 2018-04-26
Check array arguments

GomZik:
Refs #496


GomZik:
My bad, will update PR soon

GomZik:
Can not find the way to check arguments and not to break core functionality. Close PR for now.

PR #498, opened 2018-04-27
Introduce black to format source code

imanhodjaev:
In this PR as @tomchristie requested to introduce black first as a source code formatting tool at https://github.com/encode/apistar/pull/442. Next PR will be with auto-formatted APIStar source and contributor docs explaining how enable automatic code formatting if needed.

Thanks.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ jinja2
 whitenoise
 
 # Testing requirements
-black
+black; python_version == '3.6'
 codecov
 flake8
 isort
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,6 +8,7 @@ jinja2
 whitenoise
 
 # Testing requirements
+black
 codecov
 flake8
 isort
--- a/scripts/format
+++ b/scripts/format
@@ -1,0 +1,10 @@
+#!/bin/sh -e
+
+export PREFIX=""
+if [ -d 'venv' ] ; then
+    export PREFIX="venv/bin/"
+fi
+
+set -x
+
+${PREFIX}black .


jgirardet:
Black is python 3.6. Will fail in a 3.5 env.

imanhodjaev:
I've been thinking if adding another file like `requirements.dev.txt` makes sense, what do you think?

tomchristie:
Strongly prefer not to.

jgirardet:
And what about use pipenv ?

imanhodjaev:
@tomchristie @jgirardet what if we change it to conditional requirement like

```
...
black; python_version == '3.6'
```

jgirardet:
https://docs.pipenv.org/basics/#general-recommendations-version-control. We should gitignore pipefile.lock and install with pipenv install --skip-lock

tomchristie:
@jgirardet - Would be happy enough to take a PR replacing `requirements.txt` with `Pipfile` and discuss from there.

jgirardet:
Ok, i'll see for it next week

jgirardet:
@tomchristie some questions about it:

audiolion:
My personal opinion, I tried Pipenv out with a lot of optimism, and found it really wanting. It just doesn't seem stable, there are just commits straight to master with little description and no vetting process done, minor releases that break things (https://github.com/pypa/pipenv/issues/1653), and a pretty terrible experience on the issue tracker (https://github.com/pypa/pipenv/issues/1137#issuecomment-352483542). Furthermore, some things just baffle me like running `pipenv install` doesn't cache and will re-run the entire dependency resolution every single time, so you sit there waiting a minute to find out nothing needed to change. And lastly, it is just so slow compared to using pip. Kenneth has pushed it super hard by preemptively stating it is the "PyPA-recommended" package manager, but its lack of some critical pieces along with the aforementioned issues would make me ashamed to show someone new to python this "recommended" package manager.

imanhodjaev:
@tomchristie @audiolion what if I try to make something workable with Pipenv and then we can assess if it worth using it?

tomchristie:
@imanhodjaev Sure thing.

PR #501, opened 2018-04-28
Add the content length to WSGI input reading

### LINKED ISSUE #489: ASyncApp + annotation on the handler function results does not work
I have the following code:
```python
import apistar.http as http
from apistar import ASyncApp, Route


def hello_world(r: http.Request) -> dict:
    print('inside hello_world')
    return {'message': 'hello world'}


routes = [
    Route('/', method='GET', handler=hello_world)
]

app = ASyncApp(routes=routes)


if __name__ == '__main__':
    app.serve('127.0.0.1', 8000, debug=True)
```
Running this simple server by ./app.py and then going
```bash
curl -X GET -i http://localhost:8000
```
results in an "infinite loop", i.e., curl just hangs and waits for an answer.
Interestingly, when I kill the curl process, the running server outputs:
```
inside hello_world
127.0.0.1 - - [25/Apr/2018 17:07:49] "GET / HTTP/1.1" 200 -
```
When I use App instead of ASyncApp, everything works without problems.

python version: 3.6.5
package versions:
apistar    0.5.10   
certifi    2018.4.16
chardet    3.0.4    
idna       2.6      
Jinja2     2.10     
MarkupSafe 1.0      
pip        10.0.1   
requests   2.18.4   
setuptools 39.0.1   
urllib3    1.22     
Werkzeug   0.14.1   
wheel      0.31.0   
whitenoise 3.3.1

upy:
This is required to stop the input from being indefinitely read, and
thereby blocking the response

### DIFF
--- a/apistar/server/adapters.py
+++ b/apistar/server/adapters.py
@@ -34,9 +34,10 @@ class ASGItoWSGIAdapter(object):
                 return_bytes.append(message.get('body', b''))
 
         async def recieve():
+            content_length = int(environ.get('CONTENT_LENGTH', 0))
             return {
                 'type': 'http.request',
-                'body': environ['wsgi.input'].read()
+                'body': environ['wsgi.input'].read(content_length)
             }
 
         try:


tomchristie:

PR #502, opened 2018-05-01
Add support for multiple query parameter with same name

yozel:
With this PR, apistar will support multiple query parameter with same name like this:

```python
# /welcome/?name=Bob&name=Alice&name=Eve
def welcome(name: typing.List[str]=None):
    if name is None:
        return {'message': 'Welcome to API Star!'}
    return {'message': 'Welcome to API Star, %s!' % ', '.join(name)}


tomchristie:

PR #508, opened 2018-05-05
Dockerize development

danigosa:
* Added scripts based on docker to launch a development environment or directly running the tests with docker
* Scripts accepts image name as parameter for launching with Py35, Py36, or Pypy3
* Docker is based on Debian image as the only limitation to work
* Some artwork tip :)

I reused a well-tested scripts we used for a long time at my team, happy to share if you consider this tools like Docker. If you feel like having this kind of scripts are interesting and accepting PRs then I would add more tools for deployment using uWSGI, Gunicorn, Uvicorn, etc. as a real production env.

I leave a couple of screenshots of what the tool does ;)

**Launch a 30 seconds repeatable development environment** 
*Script: ./docker/development/run_development.sh [image_name]*
image

**Launch a 90 seconds repeatable tests run** 
*Script: ./docker/tests/run_tests.sh  [image_name]*

image


tomchristie:
Heya! Great PR, all nice and clean, but I'm going to take a pass on docker for now.

PR #511, opened 2018-05-06
Use a list comprehension for path fields

thedrow:
This is a slightly more efficient version of the same code.
I do believe it is clearer (which is more important).

### DIFF
--- a/apistar/document.py
+++ b/apistar/document.py
@@ -136,9 +136,10 @@ class Link:
 
         # Add in path fields for any "{param}" items that don't already have
         # a corresponding path field.
-        for path_name in url_path_names:
-            if path_name not in [field.name for field in path_fields]:
-                fields += [Field(name=path_name, location='path', required=True)]
+        field_names = [field.name for field in path_fields]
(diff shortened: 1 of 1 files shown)


apistar/document.py
```diff
@@ -136,9 +136,10 @@ def __init__(self,
 
         # Add in path fields for any "{param}" items that don't already have
         # a corresponding path field.
-        for path_name in url_path_names:
-            if path_name not in [field.name for field in path_fields]:
-                fields += [Field(name=path_name, location='path', required=True)]
+        field_names = [field.name for field in path_fields]
+        fields.extend([Field(name=path_name, location='path', required=True)
```
Bogdanp: There's no need to generate a list here so you can drop the `[`s and get rid of an allocation.

thedrow: Slightly slower since extending two lists is simply replacing a pointer, but who cares. It's clearer :)

thedrow:
@tomchristie Can we please merge this?

tomchristie:
Thanks, tho I prefer the existing style here.

PR #512, opened 2018-05-06
Improve code coverage configuration

thedrow:
Right now we're measuring coverage for the test files too. This PR fixes that.
Also branch coverage should be enabled since it's useful.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,7 @@
 [run]
 branch = 1
 include=*apistar/*
+omit=*tests/*
 
 [report]
 exclude_lines =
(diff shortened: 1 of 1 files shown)


tomchristie:
Including tests in the coverage is useful, since it's possible to otherwise fail to notice test code that's not actually running. (Eg. note that we had a PR including the tests on REST framework just today... https://github.com/encode/django-rest-framework/pull/5970)

I'd prefer to leave branch coverage off for the moment. If we get up to 100% line coverage, then perhaps we could think about also turning on branch coverage and focusing on that too.

Ta!

PR #518, opened 2018-05-08
Add Unauthorized exception

gabisurita:
Adds Unauthorized (401) exceptions, which should be used instead of 403 in case of Invalid authentication. Forbidden (403) usage usually should be limited to valid authentication but invalid authorization.   

Reference
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401


gabisurita:
Closing because its a duplicate from #500

PR #521, opened 2018-05-09
Initial work on cli

tomchristie:
Adds `apistar validate <schema_file>` and `apistar docs <schema_file>`


tomchristie:
Needs refinement, but I'd rather get it pulled in now.

PR #528, opened 2018-05-10
Docs fails in ASyncApp because is not named 'static'

PeRDy:
NoReverseMatch is raised when query `/docs/` because the docs view isn't named **static** as its sync counterpart.

```python
apistar.exceptions.NoReverseMatch: Could not build url for endpoint 'static' with values ['filename']. Did you mean 'serve_static_asgi' instead?
```


tomchristie:
Good call, thanks!

PR #545, opened 2018-05-16
Small fix to variable __name__

danigosa:
* Change __name_ by __name__ (avoid AttributeError instead of the intended TypeError)


tomchristie:
Thanks!

PR #549, opened 2018-05-17
Remove parenthesis from CustomHeadersHook example

jonatasbaldin:
No need for the `()` on the class definition.


tomchristie:
Yup, thanks! 👍

PR #529, opened 2018-05-11
fix asyncapp/docs

RunningMartin:
I can't visit `docs` when i use asyncapp

PR #551, opened 2018-05-18
Fix docs endpoint for async applications

akalipetis:
The main issue was that static asset routes were not correctly added for async applications, in case the application did not have static assets themselves. 

Also, include a very simplistic test to make sure docs endpoint works.

### DIFF
--- a/apistar/server/app.py
+++ b/apistar/server/app.py
@@ -285,7 +285,7 @@ class ASyncApp(App):
         self.injector = ASyncInjector(components, initial_components)
 
     def init_staticfiles(self, static_url: str, static_dir: str=None, packages: typing.Sequence[str]=None):
-        if not static_dir:
+        if not static_dir and not packages:
             self.statics = None
         else:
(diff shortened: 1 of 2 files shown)


audiolion:
Yeah the difference between #528 and #529 was that line of code

akalipetis:
Yes, to be honest I totally missed #528 and #529  😅

This line is definitely needed, as it does not work otherwise if `static_dir` is not set. Feel free to use #529 instead if you'd like, but please include the tests in here to make sure we catch similar issues in the future 😀

tomchristie:
Apologies I wasn't paying close enough attention here.
Looks like we want to include this PR too, right? (Both or the tests, and for the pacakge/static check)

audiolion:
this PR encapsulates what #529 was doing with an added test so probably good enough to just merge this one in

tomchristie:
Great, thanks! Released as 0.5.21

PR #554, opened 2018-05-20
WIP Pipenv integration

imanhodjaev:
In this PR I am evaluating how well pipenv will work out with APIStar build/ci/test stages.


audiolion:
from a topic I was reading on this, you aren't supposed to use a `Pipfile` / `Pipfile.lock` for libraries because they force the installer to use specific hashes of dependencies and due to dependency tree rules can cause some people to not be able to install. I believe the recommendation was to have dependency information in the setup.py file

jgirardet:
You should keep pipfile, not pipfile.lock. wild card dependency are dangerous because it might break with majors release of a dependency

imanhodjaev:
Thanks for reviews, so next things to do are

1. Ignore lock file,
2. May be discuss freezing/pinning dependencies from `requirements.txt` and `setup.py`

Also have a question regarding mypy linting may be we need to suppress some warning because they are redundant and Travis checks failing because of it.


Thanks.

Makefile
```diff
@@ -0,0 +1,7 @@
+# Makefile executes first target if all is not specified
+init:
+	pip install pipenv --upgrade
+	pipenv install --dev --skip-lock
+
+test:
+	scripts/ci
```
tomchristie: I'd rather we stick with the `/scripts/install` style than introduce a `Makefile`.

Pipfile.lock
```diff
@@ -0,0 +1,235 @@
+{
```
tomchristie: I think we'd want a `Pipfile`, but not a `Pipfile.lock`

Makefile
imanhodjaev: Fair! 👍

Pipfile.lock
imanhodjaev: Got it!

Issue #555, opened 2018-05-20
Missing str check in types.Type


aachurin:
`str` check is missing here:
```python
if args[0] is None or isinstance(args[0], (bool, int, float, list)):
    raise ValidationError('Must be an object.')
```

tomchristie:
Could you (or anyone else) submit this as a pull request?

freakabcd:
see PR #559

tomchristie:
Resolved, thanks!

PR #559, opened 2018-05-23
add missing `str` to instance check (#555)

freakabcd:
Simple addition of `str` to the instance check in Type


freakabcd:
bad PR, grr.. accidentally pushed my other change to master as well. I've switched to different branches now. Please ignore this pull request.

Issue #552, opened 2018-05-19
Schema generation fails with Exception: Cannot encode item <apistar.validators.Any object>


danigosa:
Not generating schema on `validators.Any` Type?

**apistar:** 0.5.21 using ASYncApp

This is the piece of code (that works like a charm BTW):

```python
from apistar import types, validators


class ClientQueryType(types.Type):
    sql = validators.String()
    args = validators.Array(validators.Any(), allow_null=True, default=[])


class ModelQueryType(types.Type):
    filters = validators.Object(allow_null=True, default={})
    columns = validators.Array(validators.String(), allow_null=True, default=["*"])
```

```python
➜  /app git:(dev) ✗ curl -i http://localhost:8000/schema/
[04:15:12][CRITICAL] robbie.http_.microservice.hooks.error injector.py:run:106 | Error(Unknown): Request (GET http://127.0.0.1:8000/schema/) Headers ({'host': 'localhost:8000', 'user-agent': 'curl/7.59.0', 'accept': '*/*'})
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/apistar/server/app.py", line 334, in asgi_callable
    await self.injector.run_async(funcs, state)
  File "/usr/local/lib/python3.6/site-packages/apistar/server/injector.py", line 132, in run_async
    state[output_name] = func(**func_kwargs)
  File "/usr/local/lib/python3.6/site-packages/apistar/server/app.py", line 325, in asgi_callable
    await self.injector.run_async(funcs, state)
  File "/usr/local/lib/python3.6/site-packages/apistar/server/injector.py", line 132, in run_async
    state[output_name] = func(**func_kwargs)
  File "/usr/local/lib/python3.6/site-packages/apistar/server/handlers.py", line 10, in serve_schema
    content = codec.encode(app.document)
  File "/usr/local/lib/python3.6/site-packages/apistar/codecs/openapi.py", line 506, in encode
    paths = self.get_paths(document, schema_defs=schema_defs)
  File "/usr/local/lib/python3.6/site-packages/apistar/codecs/openapi.py", line 544, in get_paths
    paths[path][method] = self.get_operation(link, operation_id, tag=tag, schema_defs=schema_defs)
  File "/usr/local/lib/python3.6/site-packages/apistar/codecs/openapi.py", line 572, in get_operation
    '#/components/schemas/'
[...]

PR #561, opened 2018-05-23
allow schema generation for validators.Any type (#552)

freakabcd:
Don't bail out when an Any type is encountered.

According to my reading of the spec, `{}` is the type that represents Any. See the Quickstart section on http://json-schema.org/ and https://cswr.github.io/JsonSchema/spec/multiple_types/

### DIFF
--- a/apistar/codecs/jsonschema.py
+++ b/apistar/codecs/jsonschema.py
@@ -301,4 +301,7 @@ class JSONSchemaCodec(BaseCodec):
                 value['uniqueItems'] = item.unique_items
             return value
 
+        elif isinstance(item, validators.Any):
+            return value
+
         raise Exception('Cannot encode item %s' % item)


freakabcd:
Reading the spec again (carefully this time), especially this part (emphasis mine):

> The most basic schema is a blank JSON object, which constrains nothing, allows anything, and describes nothing:
>
>{}
>
>You can apply constraints on an instance by adding validation keywords to the schema. For example, the “type” keyword **can be used** to restrict an instance to an object, array, string, number, boolean, or null

The schema should simply be `{}` without any specification of `type` (which only allows the specified values of `[ object, array, string, number, boolean, null ]`
I'll amend the commit and push it up.

freakabcd:
Can we comment out the `pygments_css` import in `server/handlers.py` ? It is generating build failures due to flake8 `unused import`.

danigosa:
@freakabcd merge upstream, it's already gone

freakabcd:
no-change amend commit pushed to retrigger CI

tomchristie:
Redundant since https://github.com/encode/typesystem/blob/master/typesystem/fields.py now deals with validation etc.

(And correctly uses `true` for `Any`.)

PR #571, opened 2018-05-27
Make json.dumps default function raise, not return, TypeError

anlutro:
After returning a dict of data from a route, I ran into `RecursionError: maximum recursion depth exceeded in comparison` from `abc.py` and it took me a long time to figure out why: I was returning a `datetime.datetime` as part of the data, which is not JSON serializable by default.

Turns out, because the `default` method used for JSON serialization *returns* a TypeError instead of raising it, the `default` method gets called with a TypeError over and over again, recursively, eventually hiding the original exception entirely.

I don't know if there was any reason to return the TypeError instead of raising it, but it's clearly not working as intended.

The `return` has been present since https://github.com/encode/apistar/pull/400.

### DIFF
--- a/apistar/http.py
+++ b/apistar/http.py
@@ -242,4 +242,4 @@ class JSONResponse(Response):
         if isinstance(obj, types.Type):
             return dict(obj)
         error = "Object of type '%s' is not JSON serializable."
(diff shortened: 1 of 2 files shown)


tomchristie:
Thanks!

PR #576, opened 2018-06-05
Better rendering for view docstrings

aleksarias:
Adds ability to render html formatting that may be included in doc strings


tomchristie:
Yup seems valid - just wondering which other places we ought to be doing the same?

aleksarias:
Right now this is the only place I can think of. 
In the future, if we include a section that details the response of an API endpoint... we should include the feature there too.

aleksarias:
@tomchristie, what do you think about setting the width of the docs-content pane to the minimum width where pep8 length compliant lines fit within the pane without wrapping? Users would have to wrap the lines they won't to be "formatted like code" with <code> blah.. <</code> 

I've done it on one of my projects like this: 
```
.docs-content {
    min-width: 725px;
}
code {
    white-space: pre;
    font-size: 12px;
}
```

aleksarias:
@tomchristie, is this something that should be merged?
Let me know if there any other changes that should be made in order to merge.

tomchristie:

PR #588, opened 2018-06-16
Fix resource leak on ASyncStaticFiles

danigosa:
Leak occurs when printing docs, but it would likely arise when using `ASyncStaticFiles` generally, just run async server with `PYTHONASYNCIODEBUG=1` and warnings will start to show up.

Warning:

```python 
/usr/local/lib/python3.6/asyncio/coroutines.py:110: ResourceWarning: unclosed file
        <_io.BufferedReader 
          name='/usr/local/lib/python3.6/site-packages/apistar/static/fonts/fontawesome-webfont.woff'>
       return self.gen.send(None)
```

### DIFF
--- a/apistar/server/staticfiles.py
+++ b/apistar/server/staticfiles.py
@@ -89,19 +89,23 @@ class ASGIFileSession():
                 'body': b''
             })
         else:
-            chunk = await file.read(8192)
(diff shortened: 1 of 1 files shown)


tomchristie:
Thanks!

PR #597, opened 2018-07-06
Remove `return value` from __init__ in docs

shvechikov:
__init__ shouldn't return values.


tomchristie:
Thanks. This section is actually removed in the upcoming 0.6 docs. It'll get a refresh as the type system starts to get more love again as a framework-independent tool.

Issue #595, opened 2018-06-23
apistar is not working in python3.7+


RokkerRuslan:
Framework use types.NewType for create new types. For example:
`WSGIEnviron = typing.NewType('WSGIEnviron', dict)`
And used it's in annotations for handlers which in turn checked with **issubclass**:
`elif issubclass(param.annotation, types.Type):  # apistar/server/core.py line 70`

But is incorrect. types.NewType don't return a type (class), it's return a function.

Why did it work?
At first, only if second type will be created with ABCMeta.
```
class A(metaclass=ABCMeta): pass
T = NewType('T', int)
issublclass(A, T) # False
```
There was such an option that the first argument was simply not checked.

In Python3.7, the abc module has been rewritten, and the __issubclass__ become:
```
class ABCMeta(type):
    ...
    def __subclasscheck__(cls, subclass):
        """Override for issubclass(subclass, cls)."""
        return _abc_subclasscheck(cls, subclass)
    ...
```
And C impl:
```
static PyObject *
_abc__abc_subclasscheck_impl(PyObject *module, PyObject *self, PyObject *subclass)
{
    if (!PyType_Check(subclass)) { // <-- HERE
        PyErr_SetString(PyExc_TypeError, "issubclass() arg 1 must be a class");
        return NULL;
    }
```

Result, the "hello, world" from tutorial with Python3.7:

> Traceback (most recent call last):
>   File "main.py", line 13, in <module>
>     application = App(routes)
>   File "/home/r/w/python/tps/.env/lib/python3.7/site-packages/apistar/server/app.py", line 54, in __init__
>     routes = routes + self.include_extra_routes(schema_url, docs_url, static_url)
>   File "/home/r/w/python/tps/.env/lib/python3.7/site-packages/apistar/server/app.py", line 84, in include_extra_routes
>     name='static', documented=False, standalone=True
>   File "/home/r/w/python/tps/.env/lib/python3.7/site-packages/apistar/server/core.py", line 17, in __init__
>     self.link = self.generate_link(url, method, handler, self.name)
>   File "/home/r/w/python/tps/.env/lib/python3.7/site-packages/apistar/server/core.py", line 20, in generate_link
>     fields = self.generate_fields(url, method, handler)
>   File "/home/r/w/python/tps/.env/lib/python3.7/site-packages/apistar/server/core.py", line 70, in generate_fields
>     elif issubclass(param.annotation, types.Type):
[...]

PR #598, opened 2018-07-11
python 3.7 issubclass() fix

alkovpro:
Hi!

I recently ran to an error trying to start my new project using python 3.7 and apistar!
It appeared that this happened due to a new behavior of ABCMeta.__subclasscheck__ and I ask you to merge this small PR, fixing issue #595 

Regards,
Alexander

### DIFF
--- a/apistar/types.py
+++ b/apistar/types.py
@@ -50,6 +50,12 @@ class TypeMetaclass(ABCMeta):
         validators.Validator._creation_counter += 1
         return super(TypeMetaclass, cls).__new__(cls, name, bases, attrs)
 
+    def __subclasscheck__(cls, subclass):
+        try:
+            return ABCMeta.__subclasscheck__(cls, subclass)
+        except TypeError:
+            return False
+
 
 class Type(Mapping, metaclass=TypeMetaclass):
     def __init__(self, *args, **kwargs):


robinchew:
apistar will not run at all for Python 3.7 without this fix. Someone should merge it ASAP.

d0ugal:
I just tested this, while it improve things a test still fails under 3.7. So it isn't a complete fix AFAICT.

tomchristie:
Ah, dammit, let's just get this in first.
A PR adding `3.7-dev` to the travis matrix and resolving any remaining issues would still be much appreciated.

alkovpro:
Thanks)
I just ran tests and there's one that failed **(test_get_schema)** with missing _response_ section in generated shema (diff screenshot attached)
test_get_schema_diff

I'll try to fix this before this weekend. Then I'll add `3.7-dev` to the travis matrix and make another PR.

alkovpro:
PR #612 created

tomchristie:
Is anyone able to provide a link to why 3.7 requires this `__subclasscheck__` behavior?

alkovpro:
I think this link refers to the case: https://www.python.org/dev/peps/pep-0560/#backwards-compatibility-and-impact-on-users-who-don-t-use-typing

tomchristie:
Okay, I don't see any reference to `__subclasscheck__` there. Without thinking about it too much it feels like a breaking change that made it into 3.7, but hey-ho.

PR #600, opened 2018-07-18
Fix Boolean Validation: self.value => self.values

TheCulliganMan:
This fixes allow_null=True paired with allow_coerce=True boolean validation


tomchristie:
Looks like this is completely out of sync. Not sure why it's not just targeting `validators.py` which I think is where you change is. If you're still interested I think you'd want to sync to master and reissue the PR.

TheCulliganMan:
Sorry about this one.  Went to the wrong branch.

PR #602, opened 2018-07-19
Fix composite param component

songww:
It will check request body, even request method is GET.


songww:
@tomchristie  Is this can be merged?

tomchristie:

PR #603, opened 2018-07-25
Whitenoise options on App instantiation

PeRDy:
Allow apistar `App` and `ASyncApp` to configure Whitenoise instance. That is specially useful to use `autorefresh` feature of whitenoise while running in debug mode.


tomchristie:

PR #609, opened 2018-08-14
Add tox to test multiple Python 3 versions

d0ugal:
I was poking around and noticed that the tests are failing on Python 3.7. Seems it is only a 3.7 thing, but I can't see an obvious fix. I added a tox file to easily test against multiple versions - it might be useful.

I'll try and find some time to poke around more and see if I can come up with a fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 .coverage
 .git
 .mypy_cache
+.pytest_cache
+.tox
 __pycache__
 /*.egg-info
 /htmlcov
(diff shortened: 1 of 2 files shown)


d0ugal:
Oh, it looks like #598 might have the fix I wanted.

tomchristie:
Thanks fella! Gonna take a pass on tox.
I'm happy enough with local testing being against a single env, with CI to catch any other cases.

PR #557, opened 2018-05-22
Pass `definitions` and `allow_coerce` for types.Type

asduj:

PR #610, opened 2018-08-14
Provide `allow_coerce` in Type to validators

zueve:
One row BugFix:
typing.Type is support parameter allow_coerce but now don't provide it to self.validators

this patch fixed it


simon-mo:
Hi Thanks for the PR, isn't this a duplicate of #557?

zueve:
Hi! Yes it is the same

tomchristie:
Thanks - now redundant, given https://github.com/encode/apistar/pull/649

PR #621, opened 2018-09-11
Add validation for fixed-length strings.

jeamland:
The error messages returned by min_length and max_length can be a little confusing (at least cosmetically) if the requirement is actually a string of a fixed length. This patch allows the specification of a length parameter to a String validator and also applies this logic if min_length==max_length.


tomchristie:
Redundant given that we're now using https://github.com/encode/typesystem for the validation. (But I'd happily take a similar PR there - it'd still be valid)

PR #622, opened 2018-09-11
AsyncApp, check static file modified use error api.And when dev, use adapter when read should set content_length, otherwise block

ZhouBox:


ZhouBox:
whitenoise 4.0 could modify api.

tomchristie:

Issue #490, opened 2018-04-26
on_response hook don't receive the same injected dependency as on_request


pslacerda:
Edited on_error -> on_response.

For some reason some objects injected `on_request()` isn't the same injected on `on_response()`, but other components like `WSGIEnviron` are. This happened on an exception context.

```python

class AppOpticsTracerHooks:
    @staticmethod
    def on_request(dependency: Dependency, environ: WSGIEnviron) -> None:
        dependency.attr = 1
        environ['attr'] = 1

    @staticmethod
    def on_response(dependency: Dependency, host: Host,
                 request: Request, response: Response) -> Response:
         # after a exception happened on the handler
         assert environ['attr'] == 1  # ok
         assert dependency.attr == 1  # fails
```

pslacerda:
Seems that for each `injector.run()` a new initial state is provided. Maybe could the state be preserved between `injector.run()` calls in the same `App.__call__()`?

pslacerda:
Forgot the last comment. I have no idea what I was talking about.

I wanted to say to `resolve_functions()` use the actual state together with the initial state if possible. Maybe can this actually improve the performance and fix this issue? Or there are any `Component` that needs to be always reinstantiated?

tomchristie:
> For some reason some objects injected on_request() isn't the same injected on on_response(), but other components like WSGIEnviron are. This happened on an exception context.

I'd consider this a valid bug yup - they ought to get the same state, if it has already been created.

pslacerda:
Seems that it you already fixed it somewhere between this issue and today. `state` is the same between hooks invokations.

https://github.com/encode/apistar/blob/002b525a108a2fa6b63855e2ca0fab7e1909e77b/apistar/server/app.py#L197

tomchristie:
@pslacerda Great, thanks for confirming.

PR #625, opened 2018-09-19
Load components from cache when calling error hooks

Bogdanp:
This fixes an issue where `on_error` hooks don't have access to the same component instances as other hooks and handlers. In addition, this may fix other resolution issues/heisenbugs but I was mainly concerned with fixing `on_error` as any code right now that tries to clean up resources on error (eg. db connections) based on a component instance is probably incorrect.

### DIFF
--- a/apistar/server/injector.py
+++ b/apistar/server/injector.py
@@ -82,9 +82,9 @@ class Injector(BaseInjector):
         steps.append(step)
         return steps
 
-    def resolve_functions(self, funcs):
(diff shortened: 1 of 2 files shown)


Bogdanp:
I think this issue was correctly identified in https://github.com/encode/apistar/issues/490, but mistakenly characterised as "fixed". It would *appear* to be fixed by the change made to pass `state` around, but that isn't the case as you can see from my changes to the injector.

PR #630, opened 2018-10-04
Fixed typo in apistar.yml path property.

marcorichetta:
Schema.yaml should be schema.yml to match the first file created.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Let's also create a configuration file `apistar.yml`:
 
 ```yaml
 schema:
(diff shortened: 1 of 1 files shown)


tomchristie:
Have ended up with this fixed the other way around. *But* on consideration I'd be okay with us switch `schema.yaml` occurrences in the README and docs/index.md to instead be `schema.yml` throughout, since that'd match up better with the config file.Happy to take a PR for that if anyone's up for it.

PR #635, opened 2018-10-24
Cannot use os.path.join as it results in backslash in Windows, but jinja only accepts forward slashes.

vavalomi:
Cannot use os.path.join as it results in backslash in Windows, but jinja only accepts forward slashes.
https://github.com/pallets/jinja/issues/711#issuecomment-300070379


vavalomi:
The build is failing with the errors not related to this PR

tomchristie:
Got there in the end. Now resolved.

PR #639, opened 2018-11-12
OpenAPI: Support parameter references

abn:


abn:
@tomchristie - hoping that the implementation in a7f8518 is aligned to what you had in mind for the TODO. The other two commits are not related to that change but were added to handle CI failures. Happy to field any feedback.

tomchristie:
Looks good actually! It'd need bringing up to date, but otherwise I think we'd be able to have this in (apologies for the delay!)

abn:
@tomchristie no worries; I will get it rebased soon-ish.

PR #642, opened 2018-12-22
windows issue on apistar/index.html

### LINKED ISSUE #641: docs: jinja2.exceptions.TemplateNotFound: apistar\index.html
Hello,

When using both the CLI and the programmatic interface to serve or build docs from a schema, I receive a the exception `jinja2.exceptions.TemplateNotFound: apistar\index.html`.

using apistar==0.6.0 installed via pipenv on windows

fix is to not use os.path.join for building apistar\index.html but do `theme + loader.delimiter + 'index.html'` in apistar\core.py

sdementen:
fix #641

### DIFF
--- a/apistar/core.py
+++ b/apistar/core.py
@@ -147,7 +147,7 @@ def docs(schema, format=None, encoding=None, theme='apistar', schema_url=None, s
     else:
         static_url_func = static_url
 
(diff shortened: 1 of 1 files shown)


tomchristie:
Now resolved. Thanks!

PR #653, opened 2019-03-27
Update cli.py

huangsam:
It seems as if the theme parameter was not passed in. As a result, the `index.html` does not change along with the JS/CSS resources that are made at the bottom of this particular routine. Not sure if there are other issues in this function, but this fixed theming for me.


huangsam:
This PR is pretty similar to #633 - though no one has merged that particular branch yet - and it has a merge conflict at the moment.

tomchristie:
Thanks!

PR #666, opened 2019-09-19
Allow "python -m apistar.cli"

mj026:


rpkilby:
You can define a [`__main__.py` module](https://docs.python.org/3/library/__main__.html) to enable invoking the package by name `python -m apistar`.

```python
# apistar/__main__.py
from .cli import cli

if __name__ = '__main__':
    cli()
```

maerteijn:
I know, but just adding this two lines means no new file and does the job.

PR #674, opened 2020-02-18
Added tox and python 3.8 on tests

amenezes:
Included:
- tox.ini added on root directory to test apistar with python 3.6, 3.7 and 3.8
- [branch](https://en.wikipedia.org/wiki/Code_coverage#Basic_coverage_criteria) included on test report

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,6 @@
+[run]
+branch = True
+
 [report]
 exclude_lines =
     # Have to re-enable the standard pragma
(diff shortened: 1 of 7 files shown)


tomchristie:
I'd be very happy to accept the 3.8 addition yup.

I'd rather pass on the tox part for the moment. It's not impossible that we'd add tox or nox at some point, but if we do so I'd like to think about doing it consistently across a bunch of encode projects, rather than on an ad-hoc basis.

PR #675, opened 2020-04-17
fix typesystem version

ucpr:
## Overview
fixed typesystem version.

The typesystem from 0.3.0.dev0 contains disruptive changes.
For example, `typesystem.SchemaDefinitions()` cannot be used.

https://github.com/encode/apistar/blob/8015bc1b3c9f43bcf9baa8407330338224232689/apistar/schemas/jsonschema.py#L3

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ click
 jinja2
 requests
 pyyaml
-typesystem>=0.2.0
+typesystem>=0.2.0, <0.3.0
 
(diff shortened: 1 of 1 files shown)


tomchristie:
Shouldn’t be needed 0.3.0.dev0 is a prerelease, and won’t install unless you’re using pip install —pre

ucpr:
I see. Thank you for your reply.

PR #683, opened 2021-09-17
upgrade to typesystem 0.3.0

aminalaee:
Right now `apistar` has a loose dependency on `typesystem`.

This PR will support latest release of `typesystem` and also pin dependency.

PR #684, opened 2021-09-30
Update requirements.txt

l00p1n6:
typesystem.SchemaDefinitions used in apistar/schemas/jsonschema.py are removed from typesystem 0.3.0


aminalaee:
There's another PR #683 for this.

PR #686, opened 2022-02-24
[WIP] Update typesystem

majiang:


majiang:
It turns out #683 is enough.

